PR #15, opened 2018-05-29
Rename sync command to push

domenkozar:
As pointed out by @zimbatm, sync conveys it is
bidirectional, while it's really a push.


.travis.yml
```diff
@@ -6,6 +6,6 @@ env:
   global:
     - secure: "LWfVXkoJlIHBY3yF1ylD+sX3YiHPTdNZJaopxdBpct+QCaBYEQJaJjT3qIYSjubZ9gHMLktWjLWdYSjAAQPCrlKOD8LBYPtMhg7GMzuOnBS/rSv+VhdVIH4PJ0KISZyba7QAPsDnNGm4xHlw/NeUONuNacnEBllC6+fiiEEtqtzYlRkNI71GCiSRO899uvp3/Ff+LYoY9XIgrGoRHioftoQ0GZs2K8GxhWo/UzshjqS6bV3nuiZQZH8iWC8gAAZPTK6z3dyk2xi6LFvpcOPu5wCIdjSQy0bPjCWIuQkYcJF+ql8oCfyVlkD4vAyMOb7mj+6SyhKZZ0WFhRT0tWc8Gv5PXDQ8hWrDQTtI9dPXfoB19t8BRrxRMd3vUFf/AfItcAHBX9LI8OkkSvX9MGn2hLk8f4Rz4EFJJbQDn7GIwj3pesk5hQsUizSq5jtF+MDAda0ojMGSGV8nlfBxDuYat4IdiKsi1bJTnhfZ5m1cD1yWgkOGxi77z8b8ItiWslCD7f233m5jk9464jE+BKrHOwRaEdx3Au5e7JZmJx2AtxdKcagMX6kLmmorML3odaDvkWiMYcd+ztmFDvjiebW0D96Z2IQzsTlfGb3M6PP5+vMPCw5cpYBm3O69DEGU2PI6jpvTSxrdN6lVTJL00Od9Hn9KffDNVgFZ5ijoJnTO/8A="
 script:
- - nix-env -if https://github.com/cachix/cachix/tarball/master --substituters https://cachix.cachix.org --trusted-public-keys cachix.cachix.org-1:eWNHQldwUO7G2VkjpnjDbWwy4KQ/HNxht7H4SSoMckM=
+ - nix-env -if https://github.com/cachix/cachix/tarball/master --substituters 'https://cache.nixos.org https://cachix.cachix.org' --trusted-public-keys cachix.cachix.org-1:eWNHQldwUO7G2VkjpnjDbWwy4KQ/HNxht7H4SSoMckM=
```
zimbatm: it would be cool if a cache could be configured to pull and re-sign another cache

domenkozar: For any content in another cache or just paths pushed?

domenkozar: I don't think either would work:

- it can't resign since signing key is never revealed to cachix (although you could optionally do that)
- it can't proxy since public key for domain won't match

zimbatm: Optionally `cachix use` could set both public keys

domenkozar: Thanks - opened https://github.com/cachix/cachix/issues/16

PR #31, opened 2018-06-04
Add logout api endpoint

domenkozar:


cachix-api/src/Cachix/Api.hs
```diff
@@ -68,7 +68,13 @@ data BinaryCacheAPI route = BinaryCacheAPI
   } deriving Generic
 
 data CachixAPI route = CachixAPI
-   { login :: route :-
+   { logout :: route :-
+       "logout" :>
+       CachixAuth :>
+       Post302 '[JSON] '[ Header "Set-Cookie" SetCookie
+                        , Header "Set-Cookie" SetCookie
```
puffnfresh: Why twice?

domenkozar: servant-auth always sets also csrf cookie alongside session.

PR #37, opened 2018-06-07
Implement narinfo lookup before pushing

### LINKED ISSUE #28: Optimise packages upload
I noticed that uploading a package that already exists is as slow as uploading it for the first time.

It is very noticeable  on big packages, such as `llvm` or `ghc`:  every time I call `nix-build -A xxx | cachix push name` it seems to re-upload all the dependencies.

I would expect that if the package hash already exists in cache then `cachix` could skip uploading it and move on to the next package.

domenkozar:
- Adds HEAD narinfo support for the api

fixes #28


domenkozar:
Leaving this around until tomorrow and then merging.

PR #43, opened 2018-06-11
Remove dependency on servant-auth-server

domenkozar:


domenkozar:
Blocked in https://github.com/haskell-servant/servant-auth/issues/58

domenkozar:
Depends on #149 due to http-api-data-0.3.10

domenkozar:
It's worth moving `FromJWT, ToJWT` into servant-auth, then all of this becomes trivial.

domenkozar:
Depends on https://github.com/haskell-servant/servant-auth/issues/135

PR #51, opened 2018-06-13
Travis: Remove extra `cachix push`

asymmetric:
We already start cachix in daemon mode, which watches `/nix/store` and pushes store paths, so we don't need to push again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,4 +16,4 @@ script:
   cachix.cachix.org-1:eWNHQldwUO7G2VkjpnjDbWwy4KQ/HNxht7H4SSoMckM='
 - cachix use cachix
 - cachix push cachix --watch-store&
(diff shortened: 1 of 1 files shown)


domenkozar:
There's a potential racing condition when the build finishes before daemon pushes just built path (imagine it has 1G of size). Ideally we'd wait for daemon to finish, but couldn't think of a simple way to do that. In either case `cachix push` should be a fast operation.

asymmetric:
Makes sense.

asymmetric:
I guess a way to solve this could be:

* Add signal handling to `cachix`
* Start the `cachix` daemon in the foreground as a parallel job
* in an `after_success` step, send the signal to `cachix` telling it to wrap up
* The daemon will exit once it's done

domenkozar:
Sounds reasonable, could you open an issue?

Issue #55, opened 2018-06-14
xz compression caps upload at ~3-5mbit/s


nh2:
`cachix push` is uploading with 3-5 Mbit's from my 1000 Mbit/s connection from Zurich.

I suspect the biggest reason is lack of HTTP keepalive; every upload is a new HTTP(s!) / TCP connection. That resets the TCP window size for every package.

I think that if you move this code that creates the HTTP manager

https://github.com/cachix/cachix/blob/fed0a719e4c63a693b6a593004b7a6e7e92216e8/cachix/src/Cachix/Client/Commands.hs#L235-L237

out of the `go` loop, it will use keepalive across all packages, one TCP connection, and be a lot faster.

---

Further consider to enable BBR congestion control on the server side. For my servers it has a huge impact and allows me to push 1 Gbit/s over oceans without problems.

```nix
{
    boot.kernelModules = [ "tcp_bbr" ];

    # Enable BBR congestion control
    boot.kernel.sysctl."net.ipv4.tcp_congestion_control" = "bbr";
    boot.kernel.sysctl."net.core.default_qdisc" = "fq"; # see https://news.ycombinator.com/item?id=14814530
}
```

danbst:
This may be also a throttling from server side

nh2:
I thought up some garbage here, sorry. Of course updating the pure `baseUrl` inside `env` has no effect on the `Manager` contained within `env`. Sorry for the confusion. The issue must be somewhere else.

nh2:
There is definitely something funny going on. Even changing from `mapPool 4` to `mapPool 400` doesn't change the throughput. But at closer look, `cachix` is CPU-bound. It must be doing something expensive somewhere.

domenkozar:
I'd say it's lzma compression with level 6.

domenkozar:
https://goo.gl/images/MHFiHN

nh2:
> I'd say it's lzma compression with level 6.

Oh yeah that makes sense. Are other compressions also supported? With LZMA it will be impossible to reach gigabit line speed.

nh2:
Indeed switching from `.| compress Nothing` to `.| compress (Just 0)` brings me up to ~50 Mbit/s. Better, but still factor 20 below nice working conditions.

Are there other crompression methods available?

For example, `gzip -1` can do 25 MB/s per core, `lzop` can do 400 MB/s per core.

domenkozar:
[...]

PR #57, opened 2018-06-15
Share HTTP manager for upload. Should help with #55.

nh2:
This allows to keep the connection open across requests,
so that the TCP window has a chance to grow to grow
to the full throughput of the connection.


nh2:
Sorry what I wrote here is nonsensical, moving pure code around and hoping for a change in HTTP behaviour. The manager in the `env` is already shared.

PR #82, opened 2018-07-01
Add stack2nix updating script

domenkozar:


domenkozar:
I really don't want to also build stack2nix, so this is blocked on https://github.com/input-output-hk/stack2nix/issues/94

domenkozar:
Sadly stack -> nix conversion is very impure so committing these files to git is the only way :)

roberth:
It's ok

domenkozar:
Exactly. I'm thinking of dropping/ignoring cabal files now that nixpkgs handles that better since 0f1ebc385d1a6a709900b6893901d6d914fdd5ae

Issue #120, opened 2018-08-30
narinfo corrupt?


dtzWill:
Path in question: https://allvm.cachix.org/rcdwq1bsh5lwbvcm6x01fxx5nph90nc6.narinfo

For example:

```
$  nix path-info --json /nix/store/rcdwq1bsh5lwbvcm6x01fxx5nph90nc6-source --store https://allvm.cachix.org
error: NAR info file 'rcdwq1bsh5lwbvcm6x01fxx5nph90nc6.narinfo' is corrupt
```

The path is apparently a git clone of my [NUR repo](https://github.com/dtzWill/nur-packages) FWIW.

Which is problematic for a few reasons:

* How does a corrupt narinfo get created/uploaded? Should this be checked server-side?
* How can I fix this? Did I cause it? I had to disable use of the cache when that path is involved since nix aborts everything when it finds an invalid narinfo, understandably :).

Thanks!

domenkozar:
Looking into it.

domenkozar:
@dtzWill do you have this store path around? Can you run `nix-store --dump /nix/store/rcdwq1bsh5lwbvcm6x01fxx5nph90nc6-source|wc -c`

domenkozar:
> How does a corrupt narinfo get created/uploaded?

Signature is calculated and verified. Maybe Nix doesn't accept files that have narSize of 0. Let's first verify that uploaded file is the right one.

domenkozar:
For reference I'll post narinfo here:

```
StorePath: /nix/store/rcdwq1bsh5lwbvcm6x01fxx5nph90nc6-source
URL: nar/0040f94d11d0039505328a90b2ff48968db873e9e7967307631bf40ef5679275.nar.xz
Compression: xz
FileHash: sha256:0040f94d11d0039505328a90b2ff48968db873e9e7967307631bf40ef5679275
FileSize: 32
NarHash: sha256:0mdqa9w1p6cmli6976v4wi0sw9r4p5prkj7lzfd1877wk11c9c73
NarSize: 0
Deriver: unknown-deriver
Sig: allvm.cachix.org-1:qtW0r7Fnq5uYmN7IziKk6tbrsLL+zaUFlaB46O/j7c3Q8329zgPuzN8DMPPlz3C5olOnw1oGjxdbBFw93Xk5BQ==
```

domenkozar:
[...]

PR #121, opened 2018-08-30
fingerprint now fails on narSize == 0

domenkozar:
See #120


cachix/src/Cachix/Client/Commands.hs
```diff
@@ -280,6 +281,11 @@ pushStorePath env config name storePath = do
               Right NoContent -> return ()
 
 
+-- TODO: should one error abort the whole pushing process? Or do a retry? Or keep going and then fail?
+eitherToPanic :: Show err => Either err val -> val
```
roberth: You don't need an exception in pure code to do this. Three lines up on the use site you can throw a proper IO exception

cachix-api/src/Cachix/Api/Signing.hs
```diff
@@ -18,11 +19,15 @@ import qualified Data.Text as T
...
+data FingerprintInvalid =
+  NarSizeIsZero
+  deriving (Show)
+
 -- perl/lib/Nix/Manifest.pm:fingerprintPath
--- TODO: Either Text ByteString: assert values
 -- NB: references must be sorted
-fingerprint :: Text -> Text -> Int -> [Text] -> ByteString
-fingerprint storePath narHash narSize references = toS $ T.intercalate ";"
+fingerprint :: Text -> Text -> Int -> [Text] -> Either FingerprintInvalid ByteString
```
roberth: Not sure why fingerprinting should fail. It's the narinfo that's invalid, not that the fingerprint can't be made. Isn't there someplace else where you could also validate the store path, hash format etc?

PR #123, opened 2018-09-01
Allow specifying path to the config file on cli

domenkozar:
To achieve that, factor out all logic of setting up environment into `mkEnv :: CachixOptions -> IO Env`

### DIFF
--- a/cachix/src/Cachix/Client/Commands.hs
+++ b/cachix/src/Cachix/Client/Commands.hs
@@ -50,6 +50,8 @@ import           Cachix.Client.Config           ( Config(..)
                                                 , mkConfig
                                                 )
 import qualified Cachix.Client.Config          as Config
+import           Cachix.Client.Env              ( Env(..) )
+import           Cachix.Client.OptionsParser    ( CachixOptions(..) )
 import           Cachix.Client.InstallationMode
 import           Cachix.Client.NixVersion       ( getNixVersion
                                                 , NixVersion
@@ -64,10 +66,10 @@ cachixClient = fromServant $ client Api.servantApi
 cachixBCClient :: Text -> Api.BinaryCacheAPI (AsClientT ClientM)
 cachixBCClient name = fromServant $ Api.cache cachixClient name
 
-authtoken :: ClientEnv -> Maybe Config -> Text -> IO ()
-authtoken _ maybeConfig token = do
+authtoken :: Env -> Text -> IO ()
+authtoken env token = do
   -- TODO: check that token actually authenticates!
-  writeConfig $ case maybeConfig of
+  writeConfig (configPath (cachixoptions env)) $ case config env of
     Just config -> config { authToken = token }
     Nothing -> mkConfig token
   putStrLn ([hereLit|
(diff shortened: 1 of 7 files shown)


domenkozar:
Dhall needs absolute path to work, so I need to fix that.

cachix/src/Cachix/Client/Commands.hs
```diff
@@ -78,21 +80,21 @@ Continue by creating a binary cache with:
 and share it with others over https://<name>.cachix.org
   |] :: Text)
 
-create :: ClientEnv -> Maybe Config -> Text -> IO ()
-create _ Nothing _ = throwIO $ NoConfig "start with: $ cachix authtoken <token>"
-create env (Just config@Config{..}) name = do
+create :: Env -> Text -> IO ()
+create (Env Nothing _ _) _ = throwIO $ NoConfig "start with: $ cachix authtoken <token>"
```
roberth: This should be `Env { config = Nothing }`. Imho we should avoid positional matching except in newtypes and single-field sums.

cachix/src/Cachix/Client/Commands.hs
```diff
@@ -111,10 +113,10 @@ To instruct Nix to use the binary cache:
 
       |] :: Text)
 
-use :: ClientEnv -> Maybe Config -> Text -> Bool -> IO ()
-use env _ name shouldEchoNixOS = do
+use :: Env -> Text -> Bool -> IO ()
+use Env{..} name shouldEchoNixOS = do
```
roberth: I see you're still doing this. The least you could do is use namedfieldpuns. It's still name shadowing, but at least we'll have a clear declaration site for the variable.

cachix/src/Cachix/Client/Config.hs
```diff
@@ -35,24 +36,23 @@ mkConfig authtoken = Config
...
   then Just <$> input auto (toS filename)
   else return Nothing
 
-
-getFilename :: IO FilePath
-getFilename = do
+getDefaultFilename :: IO FilePath
+getDefaultFilename = do
   dir <- getXdgDirectory XdgConfig "cachix"
   createDirectoryIfMissing True dir
```
roberth: Side effect in something IO action called `get...`.

cachix/src/Cachix/Client/OptionsParser.hs
```diff
@@ -13,21 +13,30 @@ import URI.ByteString            (URIRef, Absolute, parseURI, strictURIParserOpt
...
 data CachixOptions = CachixOptions
   { host :: URIRef Absolute
+  , configPath :: Config.ConfigPath
   , verbose :: Bool
   } deriving Show
 
-parserCachixOptions :: Parser CachixOptions
-parserCachixOptions = CachixOptions
+parserCachixOptions :: Config.ConfigPath -> Parser CachixOptions
+parserCachixOptions configpath = CachixOptions
```
roberth: s/configpath/defaultConfigPath

cachix/src/Cachix/Client/Commands.hs
domenkozar: Ah, I didn't know you can pattern match like that! Thanks

cachix/src/Cachix/Client/Config.hs
domenkozar: Just a line below there is `getXdgDirectory` - what's wrong with this naming?

cachix/src/Cachix/Client/Commands.hs
domenkozar: Guilty as charged :) I wanted to refactor the code without too many changes - but somehow this backfired by disobeying the boy scout rule. I'll convert those to plain function calls.

cachix/src/Cachix/Client/Config.hs
domenkozar: Ah! That's a valid point.

PR #133, opened 2018-09-26
Warnings

domenkozar:


cachix-api/src/Cachix/Api/Swagger.hs
```diff
@@ -1,3 +1,4 @@
+{-# OPTIONS_GHC -fno-warn-orphans #-}
```
roberth: Orphans should go in their own file to prevent this flag from masking unintended orphans. Also explicit is good.

domenkozar: This is the file that only contains orphans.

roberth: Hehe ok. I would call it `Cachix.Api.Swagger.Orphans`, so you won't accidentally add non-orphan things to it.

PR #142, opened 2018-10-07
Ditch hpack for Cabal common stanzas

domenkozar:
I'm done dealing with hpack vs cabal out of sync issues


roberth:
Oops, didn't notice that the relation with the other PR which approves this.

PR #185, opened 2019-03-19
Allow exitSuccess

roberth:
This is required for command line completions to succeed.


domenkozar:
I'm not familiar with `fromException` semantics, will this also catch CachixException?

roberth:
It's based on `Typeable` and will only match `ExitCode` exceptions, typically produced by `exitWith` and friends.

PR #187, opened 2019-03-21
netrc: create possibly missing directories

domenkozar:
Avoids 

```
$ cachix use private
/home/ielectric/.config/nix/netrc: openBinaryFile: does not exist (No such file or directory)
```

### DIFF
--- a/cachix/src/Cachix/Client/NetRc.hs
+++ b/cachix/src/Cachix/Client/NetRc.hs
@@ -8,7 +8,8 @@ import           Data.List            (nubBy)
 import qualified Data.Text          as T
 import           Network.NetRc
 import           Protolude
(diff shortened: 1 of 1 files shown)


cachix/src/Cachix/Client/NetRc.hs
```diff
@@ -8,7 +8,8 @@ import           Data.List            (nubBy)
 import qualified Data.Text          as T
 import           Network.NetRc
 import           Protolude
-import           System.Directory     ( doesFileExist )
+import           System.FilePath.Posix ( takeDirectory )
```
roberth:
```suggestion
import           System.FilePath       ( takeDirectory )
```

PR #193, opened 2019-05-29
Export keys

roberth:
- [x] rebase

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,13 +20,15 @@ Available options:
   -h,--help                Show this help text
   -h,--host URI            Host to connect to (default: https://cachix.org)
   -c,--config CONFIGPATH   Cachix configuration
-                           file (default: "/home/ielectric/.config/cachix/cachix.dhall")
+                           file (default: "~/.config/cachix/cachix.dhall")
   -v,--verbose             Verbose mode
   -V,--version             Show cachix version
 
 Available commands:
   authtoken                Configure token for authentication to cachix.org
-  create                   Create a new binary cache
+  export                   Read local secrets and configuration and export them
+                           in JSON Lines format
+  generate-keypair         Generate keypair for an binary cache
   push                     Upload Nix store paths to the binary cache
   use                      Configure nix.conf to enable binary cache during
                            builds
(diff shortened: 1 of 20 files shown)


domenkozar:
TLDR: I'm going to argue that JSONL is not the best choice for export and we should rather use plain JSON.

Currently, the format looks like:

```
{"kind":"CachixSigningKey","cacheName":"hercstg-private","secretKey":"DummySecretKey="}
{"kind":"CachixPullToken","cacheName":"hercstg-private","secretToken":"DummySecretToken"}
{"kind":"CachixPublicKey","cacheName":"hercstg-private","publicKey":"hercstg-private.cachix.org-1:12313233453564534245445="}
```

While I propose to normalize and drop the lines addition to:

```
{
  "hercstg-private": { "authToken": "Dummy", "publicKeys": ["Dummy"], "signingKey": "Dummy" },
  "hercstg-private2": { "authToken": "Dummy2", "publicKeys": ["Dummy2"], "signingKey": "Dummy2" } 
  ...
}
```

# Easier parsing

JSON Lines is not a standard yet (first time I hear about it in my not-so-short career),
so some tools might support it (like jq) but most won't. So really you need to do CSV + JSON parsing, which is additional work.

That makes total sense as JSONL documentation states to use lines instead of an array when streaming, but that's not a requirement for cachix export.

So parsing is really just JSON.

# More work to write it by hand (normalization)

It's quite tedious to write this by hand. Best way to count would be per extra character (excluding things that are copy pasted like all the keys).

It's also redundant to write `CachixPublicKey` and then `publicKey` while repeating `cacheName` for each entry.

## Why do we need to write export format by hand?

The format will be used by Hercules CI Agent, some things in the future might come from the web (like cache specific tokens). 

A bit more practical reason (UX-wise) is to not blindly trust the export output, as those secrets could be wrong. Even for security reasons, the user shouldn't have to do this locally, or they'll also have to delete the data as an additional step to exporting.

# Human readable

This one adds a bit on top of the previous point, I'd argue the export needs to be human readable for auditing. Proposed format groups cache specific fields and removes redundancy, making it easier for humans to read.

roberth:
We'll reconsider a variation of this feature at a later time.

PR #194, opened 2019-06-01
Remove ambiguous, shadowed -h for host option

roberth:


domenkozar:
That should actually be an optparse-applicative error

roberth:
Should probably be covered with a generic test. Doing this in types is obnoxious for little gain.

PR #195, opened 2019-06-05
Support servant-0.16

domenkozar:


domenkozar:
Isn't that already the case given that `Servant.API` is imported and `Cachix.Servant`, otherwise it would conflict?

PR #197, opened 2019-06-10
Add shell autocompletion

roberth:


domenkozar:
Is there a way to hint that with Cabal or similar? Most users install cachix via nixpkgs, which won't have this change.

roberth:
Nope, I authored the function and such a Cabal feature would make me happy but there isn't one. Seems to be a matter of adding it to nixpkgs like some of the other haskell tools. (Cachix didn't support it at that time, so it probably still needs to be done)

domenkozar:
I guess cabal2nix could detect optparse-applicative in executable dependencies and add the flag, but that's for another day

roberth:
Yeah, such an approach was also discussed here https://github.com/NixOS/nixpkgs/pull/42657

PR #215, opened 2019-08-14
LTS-14.0 and haskell.nix

domenkozar:


nix/packages.nix
```diff
@@ -0,0 +1,3 @@
+{ writeScriptBin
+}:
+{ foo = writeScriptBin "foo" "echo foo" ; }
```
roberth: Use it or lose it :)

nix/default.nix
```diff
@@ -0,0 +1,9 @@
+{ sources ? import ./sources.nix }:
+with
+  { overlay = _: pkgs:
+      { inherit (import sources.niv {}) niv;
+        packages = pkgs.callPackages ./packages.nix {};
```
roberth: Perhaps `pkgs.cachixPackages` or `pkgs.projectPackages`.

default.nix
```diff
@@ -1,24 +1,20 @@
...
-    overrides = self: super: {
-      cachix = addBuildDepends (generateOptparseApplicativeCompletion "cachix" (justStaticExecutables (filterStack super.cachix))) [ pkgs.boost ];
-      cachix-api = filterStack super.cachix-api;
-      # TODO: get stack2nix/cabal2nix to fill these in
-      inherit (pkgs.darwin.apple_sdk.frameworks) Cocoa CoreServices;
-    };
+  sources = (import ./nix/sources.nix);
+  pkgs = import sources.nixpkgs { overlays = [(self: super: { nix-store = self.nix; nix-main = self.nix; })];} ;
+  haskell = import (sources."haskell.nix") { inherit pkgs; };
+  gitignoreSource = (import sources.gitignore { inherit (pkgs) lib; }).gitignoreSource;
```
roberth:
```suggestion
  inherit (import sources.gitignore { inherit (pkgs) lib; })
    gitignoreSource;
```

default.nix
```diff
@@ -1,24 +1,20 @@
...
-    }
-  );
-  hsPkgs = (import ./stack2nix.nix { inherit pkgs; }).override {
-    overrides = self: super: {
-      cachix = addBuildDepends (generateOptparseApplicativeCompletion "cachix" (justStaticExecutables (filterStack super.cachix))) [ pkgs.boost ];
-      cachix-api = filterStack super.cachix-api;
-      # TODO: get stack2nix/cabal2nix to fill these in
-      inherit (pkgs.darwin.apple_sdk.frameworks) Cocoa CoreServices;
-    };
+  sources = (import ./nix/sources.nix);
```
roberth:
```suggestion
  sources = import ./nix/sources.nix;
```

default.nix
```diff
@@ -1,24 +1,20 @@
...
+  pkgs = import sources.nixpkgs { overlays = [(self: super: { nix-store = self.nix; nix-main = self.nix; })];} ;
+  haskell = import (sources."haskell.nix") { inherit pkgs; };
```
roberth:
```suggestion
  haskell = import sources."haskell.nix" { inherit pkgs; };
```

default.nix
```diff
@@ -1,24 +1,21 @@
...
 let
-  filterStack = drv: overrideCabal drv (super: {
-     src = builtins.filterSource
-      (path: type: baseNameOf path != ".stack-work")
-      super.src;
-    }
-  );
-  hsPkgs = (import ./stack2nix.nix { inherit pkgs; }).override {
-    overrides = self: super: {
-      cachix = addBuildDepends (generateOptparseApplicativeCompletion "cachix" (justStaticExecutables (filterStack super.cachix))) [ pkgs.boost ];
```
roberth: Can we enable
 - static
 - optparse applicative completions?

domenkozar: You don't need static because haskell.nix builds each component so executable is already statically compiled haskell wise.

I'm not sure optparse applicative completion exists for haskell.nix yet, but it doesn't matter since most folks should have cachix installed via nixpkgs.

Issue #201, opened 2019-07-04
Mass query push


domenkozar:
Currently when syncing a huge closure, for each derivation we need to do three http requests:

a) look up narinfo
b) compress + send nar
c) upload nar

For evaluations with a lot of derivations there's a significant overhead. (a) and (c) can be batched, while (b) is going to be capped at CPU (compression) and bandwidth limits.

PR #216, opened 2019-08-14
Add bulk store path query

domenkozar:
refs #201

### DIFF
--- a/cachix-api/src/Cachix/Api.hs
+++ b/cachix-api/src/Cachix/Api.hs
@@ -69,6 +69,12 @@ data BinaryCacheAPI route = BinaryCacheAPI
       CachixAuth :>
       Capture "narinfo" NarInfoC :>
       Head
+  , narinfoBulk :: route :-
+      CachixAuth :>
+      "narinfo" :>
+      Summary "Given a list of store hashes, return a list of those that are missing" :>
+      ReqBody '[JSON] [Text] :>
+      Post '[JSON] [Text]
   , createNarinfo :: route :-
       Capture "narinfo" NarInfoC :>
       ReqBody '[JSON] NarInfoCreate.NarInfoCreate :>
--- a/cachix/src/Cachix/Client/Push.hs
+++ b/cachix/src/Cachix/Client/Push.hs
@@ -55,7 +55,9 @@ data PushCache = PushCache
   , pushCacheToken :: Token
   }
 data PushStrategy m r = PushStrategy
-  { onAttempt :: RetryStatus -> m ()
+  { onAlreadyPresent :: m r -- ^ Called when a path is already in the cache.
+  , onAttempt :: RetryStatus -> m ()
+  , on401 :: m r
   , onError :: ClientError -> m r
   , onDone :: m r
   , withXzipCompressor :: forall a. (ConduitM ByteString ByteString (ResourceT IO) () -> m a) -> m a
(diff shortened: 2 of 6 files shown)


cachix/src/Cachix/Client/Push.hs
```diff
@@ -77,95 +76,81 @@ pushSingleStorePath
   -> Text -- ^ store path
   -> m r -- ^ r is determined by the 'PushStrategy'
 pushSingleStorePath ce _store cache cb storePath = retryPath $ \retrystatus -> do
-
   let (storeHash, storeSuffix) = splitStorePath $ toS storePath
       name = pushCacheName cache
```
roberth: If you remove the presence check from `pushSingleStorePath`, doesn't that make cachix overwrite store paths, for example when using `--watch-store`?
We may have to split this function into one that checks (old name) and one that doesn't.

cachix-api/src/Cachix/Api.hs
```diff
@@ -69,6 +69,12 @@ data BinaryCacheAPI route = BinaryCacheAPI
       CachixAuth :>
       Capture "narinfo" NarInfoC :>
       Head
+  , narinfoBulk :: route :-
+      CachixAuth :>
+      "narinfo" :>
+      Summary "Given a list of store hashes, return a list of those that are missing" :>
+      ReqBody '[JSON] [Text] :>
+      Get '[JSON] [Text]
```
roberth: Although `ReqBody` + `Get` seems very sensible from a REST view, the real world may disagree.
I'd prefer to configure an HTTP cache to cache a POST properly over having to debug one because we're using a corner case.

cachix/src/Cachix/Client/Push.hs
domenkozar: Ugh, need to fix that.

cachix-api/src/Cachix/Api.hs
domenkozar: Not sure what do you mean with `having to debug`, is that due to HTTP GET caching? All cachix APIs are not cached atm so there shouldn't be any issues.

roberth: I was trying to say that GET requests with request bodies are not really standard. They are allowed by the spec but [apparently](https://stackoverflow.com/questions/978061/http-get-with-request-body/983458#983458) they shouldn't influence the response, making them useless.
So yeah we may run into caching issues if we add it ourselves, or if someone decided to proxy it locally.

domenkozar: There's an update edit at the bottom. From RFCs published in 2014 body restriction was removed.

However, there's going to be very little caching involved here so better save than sorry.

cachix/src/Cachix/Client/Push.hs
```diff
@@ -76,102 +81,104 @@ pushSingleStorePath
...
+            escalate $ Api.isNarInfoCreateValid nic
+
+            -- Upload narinfo with signature
+            escalate <=< (`runClientM` clientEnv) $ Api.createNarinfo
+              (cachixBCClient name)
+              (Api.NarInfoC storeHash)
+              nic
+    onDone cb
+
+-- Retry up to 5 times for each store path.
```
roberth: Why is it 5 instead of 3? Do we need to specify it in this comment?
It should probably be a haddock.
This function is not only used for paths, so the name should change.

domenkozar:
rebased

Issue #186, opened 2019-03-21
Setting up private cache on NixOS with untrusted user fails


domenkozar:
# error 

```
$ cachix use foo
/etc/nixos/cachix: createDirectory: permission denied (Permission denied)
```

# improvements

- note which method is being used in the output
- fix the bug?

mynameisdaniil:
Similar problem on Ubuntu 18.04
`elm init` or `elm repl` produces
`elm: /usr/lib/node_modules: createDirectory: permission denied (Permission denied)`

domenkozar:
I don't think elm has much to do with cachix :)

arianvp:
Yes love to have this fixed. I don't use `/etc/nixos` to store my nixos config. I'd much rather have the following output:

```
[arian@t430s:~/Projects/generics-mrsop-diff]$  cachix use arianvp
Hey I see you're using NixOS and not running root. Please add the following to your NixOS config
{
  nix = {
    binaryCaches = [
      "https://arianvp.cachix.org"
    ];
    binaryCachePublicKeys = [
      "arianvp.cachix.org-1:/NYL/rC71vauTeFWVMJXXadA/2MfziDb+/1DlLLXUvw="
    ];
    
  };
}

```

domenkozar:
There is `-d,--nixos-folder ARG    Base directory for NixOS  configuration (default: "/etc/nixos/")`, would that suffice to write it out wherever is your configuration?

Rizary:
Is this the same with error `/etc/nixos/cachix: openFile: permission denied (Permission denied)` ? Because I have this error when using cachix in WSL ubuntu 18.04

domenkozar:
Are you using NixOS or plain Nix?

Rizary:
I am using plain Nix, in ubuntu 18.04 WSL (windows 10) version.

domenkozar:
Does `/etc/NIXOS` or `/etc/nixos` exists?

Rizary:
no it does not.

domenkozar:
@Rizary what exact command are you running?

Rizary:
@domenkozar 
`cachix use all-hies`

domenkozar:
@Rizary did you manage to resolve this? I still don't understand how you end up in that code path.

PR #222, opened 2019-09-02
Clean use command

domenkozar:
Fixes https://github.com/cachix/cachix/issues/186

- drop support for Nix 2.0.1 or lower
- install nixos only with root
- display human error if /etc/nixos is not writabble
- pretty print only cachix exceptions

### DIFF
--- a/cachix/src/Cachix/Client/InstallationMode.hs
+++ b/cachix/src/Cachix/Client/InstallationMode.hs
@@ -6,7 +6,10 @@ module Cachix.Client.InstallationMode
     getInstallationMode,
     addBinaryCache,
     isTrustedUser,
-    getUser
+    getUser,
+    fromString,
+    toString,
+    UseOptions (..)
     )
 where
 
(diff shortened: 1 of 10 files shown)


cachix/src/Cachix/Client/OptionsParser.hs
```diff
@@ -127,11 +120,13 @@ parserCachixCommand =
...
-                <$> switch
-                      ( long "nixos"
-                          <> short 'n'
-                          <> help "Write NixOS modules"
+        <*> ( InstallationMode.UseOptions
+                <$> optional
+                      ( option (maybeReader InstallationMode.fromString)
+                          ( long "mode"
+                              <> short 'm'
+                              <> help "Mode in which to configure binary caches for Nix. Supported values: nixos, root-nixconf, user-nixconf"
```
roberth: This list should be generated using `Bounded` and `Enum` instances.

cachix/src/Cachix/Client/InstallationMode.hs
```diff
@@ -102,8 +120,13 @@ nixosBinaryCache maybeConfig maybeUser bc@Api.BinaryCache {..} UseOptions {useNi
     glueModuleFile = toplevel <> ".nix"
     cacheModuleFile :: Text
     cacheModuleFile = toplevel <> "/" <> toS name <> ".nix"
-    extra :: Text
-    extra = maybe "" (\user -> [i|trustedUsers = [ "root" "${user}" ];|]) maybeUser
+    deadendInstructions :: Text
```
roberth: Suggestion: `noEtcPermissionInstructions`

domenkozar:
> This is not refactoring but cleanup + bugfix. The branch and PR names are wrong.

Started as a refactoring, ended up being a huge cleanup. Was too lazy to fix the naming.

cachix/src/Cachix/Client/OptionsParser.hs
domenkozar: I wanted to use that, but there were two issues:

- can't derive since it has more than one constructor
- we don't want to show `UntrustedRequiresSudo` as an option since that's an error

So I'll just add a comment to ADT to update this list when adding/removing stuff

roberth: :+1: 
I guess you could add a type with Enum for just this purpose, but that may not be a real improvement.

PR #227, opened 2019-09-09
Nix 2.3: bump to c++17 and depend on boost

domenkozar:
Nix 2.3 was released, these are the fixes to make `cabal2nix` output a Nix derivation that works out of the box.


domenkozar:
I've checked 2.2 support, but we really just need to be able to build with 2.3, as Cachix comes from nixpkgs which follows the latest Nix release.

PR #233, opened 2019-09-23
push: if multiple caches exist, pick the latest signing key

domenkozar:
this is common if someone generates a new key, they most probably want
to use the latest key, although we better get more sophisticated.


dmjio:
LGTM

PR #240, opened 2019-10-01
push: Don't complain on empty stdin

domenkozar:
- [ ] test with nodejs child_process

### DIFF
--- a/cachix/src/Cachix/Client/Commands.hs
+++ b/cachix/src/Cachix/Client/Commands.hs
@@ -150,17 +150,16 @@ use env name useOptions = do
       addBinaryCache (config env) binaryCache useOptions
         $ fromMaybe (getInstallationMode nixEnv) (useMode useOptions)
 
--- TODO: lots of room for perfomance improvements
+-- TODO: lots of room for performance improvements
 push :: Env -> PushArguments -> IO ()
-push env (PushPaths opts name rawPaths) = do
-  hasNoStdin <- hIsTerminalDevice stdin
-  when (not hasNoStdin && not (null rawPaths)) $ throwIO $ AmbiguousInput "You provided both stdin and store path arguments, pick only one to proceed."
+push env (PushPaths opts name cliPaths) = do
+  hasStdin <- not <$> hIsTerminalDevice stdin
   inputStorePaths <-
-    if hasNoStdin
-      then return rawPaths
-      else T.words <$> getContents
-  -- TODO: if empty, take whole nix store and warn: nix store-path --all
-  when (null inputStorePaths) $ throwIO $ NoInput "You need to specify store paths either as stdin or as a cli argument"
+    case (hasStdin, cliPaths) of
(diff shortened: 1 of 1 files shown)


cachix/src/Cachix/Client/Commands.hs
```diff
@@ -150,16 +150,20 @@ use env name useOptions = do
...
-      else T.words <$> getContents
-  -- TODO: if empty, take whole nix store and warn: nix store-path --all
+      then return ""
+      else getContents
+  when (stdinContents /= "" && not (null rawPaths)) $ throwIO
+    $ AmbiguousInput "You provided both stdin and store path cli arguments, provide only one to proceed."
+  let inputStorePaths =
+        if stdinContents == ""
+          then rawPaths
+          else T.words stdinContents
```
roberth: stdinContents handling introduces a corner case (what if `stdinContents == " "`) and isn't very clear. Ideas:
 - immediately `T.words` stdinContents
 - turn three conditionals into one by pattern matching on `(stdinContents, rawPaths) :: ([Text], [Text])`

It's also a bit weird that `rawPaths` magically turn into `inputStorePaths` which aren't raw. Probably naming can be simplified.

cachix/src/Cachix/Client/Commands.hs
```diff
@@ -150,16 +150,20 @@ use env name useOptions = do
...
   hasNoStdin <- hIsTerminalDevice stdin
-  when (not hasNoStdin && not (null rawPaths)) $ throwIO $ AmbiguousInput "You provided both stdin and store path arguments, pick only one to proceed."
-  inputStorePaths <-
+  stdinContents <-
     if hasNoStdin
-      then return rawPaths
-      else T.words <$> getContents
-  -- TODO: if empty, take whole nix store and warn: nix store-path --all
+      then return ""
+      else getContents
```
roberth:
```suggestion
      else T.words <$> getContents
```

cachix/src/Cachix/Client/Commands.hs
```diff
@@ -150,17 +150,16 @@ use env name useOptions = do
...
   inputStorePaths <-
-    if hasNoStdin
-      then return rawPaths
-      else T.words <$> getContents
-  -- TODO: if empty, take whole nix store and warn: nix store-path --all
-  when (null inputStorePaths) $ throwIO $ NoInput "You need to specify store paths either as stdin or as a cli argument"
+    case (hasStdin, cliPaths) of
+      (False, []) -> throwIO $ NoInput "You need to specify store paths either as stdin or as an command argument"
+      (True, []) -> T.words <$> getContents
+      -- if we get both stdin and cli args, prefer cli args
```
roberth:
```suggestion
      -- If we get both stdin and cli args, prefer cli args.
      -- This avoids hangs in cases where stdin is non-interactive but unused by caller
      -- some programming environments always create a (non-interactive) stdin
      -- that may or may not be written to by the caller.
      -- This is somewhat like the behavior of `cat` for example.
```

PR #241, opened 2019-10-03
travis -> github actions

domenkozar:


default.nix
```diff
@@ -14,6 +14,6 @@ let
     ];
   };
   packages = pkgSet.config.hsPkgs;
-in packages.cachix.components.exes.cachix  // {
+in packages.cachix.components.all  // {
```
roberth: I suppose this isn't public api.

domenkozar: Yes it is: https://input-output-hk.github.io/haskell.nix/reference/library/#component

roberth: Oh I meant cachix `/default.nix`

domenkozar: Kind of - need to fix readme I guess.

PR #248, opened 2019-11-01
delete cache support

domenkozar:


cachix-api/src/Cachix/Api.hs
```diff
@@ -54,6 +54,11 @@ data BinaryCacheAPI route
                :- CachixAuth
                :> ReqBody '[JSON] BinaryCacheCreate.BinaryCacheCreate
                :> Post '[JSON] NoContent,
+        delete
+          :: route 
+               :- CachixAuth -- TODO: disallow auth tokens
```
roberth:
```suggestion
               :- CachixAuth
```

I don't think we need to distinguish this in the api yet. We can add these details later when we've figured out the right way to represent it. When we do so, we need to consider all endpoints anyway, not just this one.

PR #249, opened 2019-11-11
Serve nar contents

domenkozar:
I've also factored out `NarC` into separate module and renamed it to `NarFileName`, but then decided this shouldn't use nar hash, but rather store path hash.

### DIFF
--- a/cachix-api/src/Cachix/Api.hs
+++ b/cachix-api/src/Cachix/Api.hs
@@ -116,7 +116,7 @@ data BinaryCacheStreamingAPI route
                :- "nar"
                :> StreamBody NoFraming XNixNar (ConduitT () ByteString (ResourceT IO) ())
                :> Post '[JSON] NoContent,
-        serveNar 
+        serveNarContent
           :: route
               :- CachixAuth
               :> "serve"
--- a/cachix-api/src/Cachix/Api.hs
+++ b/cachix-api/src/Cachix/Api.hs
@@ -14,11 +14,13 @@ module Cachix.Api
     BinaryCacheStreamingAPI (..),
     BinaryCachStreamingServantAPI,
     module Cachix.Api.Types,
-    module Cachix.Types.ContentTypes
+    module Cachix.Types.ContentTypes,
+    module Cachix.Types.NarFileName
     )
 where
 
 import Cachix.Api.Types
+import Cachix.Types.NarFileName (NarFileName(..))
 import qualified Cachix.Types.BinaryCacheAuthenticated as BinaryCacheAuthenticated
 import qualified Cachix.Types.BinaryCacheCreate as BinaryCacheCreate
 import Cachix.Types.ContentTypes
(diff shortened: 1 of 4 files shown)


domenkozar:
CI fails due to macos Catalina, so I guess it should be ignored for now :(

cachix-api/src/Cachix/Api.hs
```diff
@@ -106,13 +109,21 @@ data BinaryCacheStreamingAPI route
...
-               :> Capture "nar" NarC
+               :> Capture "nar" NarFileName
                :> StreamGet NoFraming OctetStream (ConduitT () ByteString (ResourceT IO) ()),
         createNar
           :: route
                :- "nar"
                :> StreamBody NoFraming XNixNar (ConduitT () ByteString (ResourceT IO) ())
-               :> Post '[JSON] NoContent
+               :> Post '[JSON] NoContent,
+        serveNar
```
roberth:
```suggestion
        serveNarContent
```

cachix-api/src/Cachix/Api.hs
```diff
@@ -106,13 +109,21 @@ data BinaryCacheStreamingAPI route
...
         createNar
           :: route
                :- "nar"
                :> StreamBody NoFraming XNixNar (ConduitT () ByteString (ResourceT IO) ())
-               :> Post '[JSON] NoContent
+               :> Post '[JSON] NoContent,
+        serveNar 
+          :: route
+              :- CachixAuth
+              :> "serve"
```
roberth: Idea:

```suggestion
              :> "nix"
              :> "store"
```

cachix-api/src/Cachix/Api.hs
```diff
@@ -106,13 +109,21 @@ data BinaryCacheStreamingAPI route
...
+              :> Capture "storepath" Text
```
roberth: Unless the implementation changed since the demo link I saw, this is actually a store hash including `.narinfo`. Will it support proper store paths?

domenkozar: It changed, it's now only a store hash. It used to be `NarInfoC`, now it's just `Text`.

roberth: Cool. If we also strip off the `-$name` we can serve working hyperlinks to store paths (!)

cachix-api/src/Cachix/Api.hs
domenkozar: I'm not  sure this is a good idea. If we ever allow changing `/nix/store` prefix, this will look weird.

It's also not entirely full path, since we're using store hash and not store path.

roberth: Just need to strip the `-$name` part to make it work based on hash, or even better is to query the real store path. It's in the narinfo.
If we allow changing the prefix, we can do a `CaptureAll` at the root and make sure it's the last route that is tried.
For security we can whitelist or blacklist prefixes. Limiting to `/nix/store` is a good start.
Serving broken links is bad ux.

roberth: This should be done separately. Resolving.

PR #264, opened 2020-01-06
Storepath size

roberth:
```
> store <- openStore
> vpi <- queryPathInfo store "/nix/store/lb3hli8d9536g45mndwfwyi6fpny0blh-bash-interactive-4.4-p23"
> validPathInfoNarSize vpi
5659704
```

### DIFF
--- a/repl
+++ b/repl
@@ -1,0 +1,8 @@
+#!/usr/bin/env nix-shell
+#!nix-shell -i bash
+
+# A Haskell repl for hacking on the cachix library
+
+# -fobject-code: required by inline-c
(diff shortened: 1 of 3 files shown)


repl
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env nix-shell
```
domenkozar: This script should go to `scripts`

roberth: Oh, `scripts/` doesn't exist yet. Are you sure? Alternatively it can go in the `cachix/` directory, which it is used for.

domenkozar: Been a while, seems ok to keep it top-level for now.

PR #265, opened 2020-01-07
Print store path size

### LINKED ISSUE #175: Calculate store size before pushing
Use System.Directory.fileSize and benchmark on huge folder and on many files.

Use case: someone accidentally wants to upload 8GB file into store would think that cachix is stuck.

domenkozar:
Fixes #175

### DIFF
--- a/cachix/src/Cachix/Client/Push.hs
+++ b/cachix/src/Cachix/Client/Push.hs
@@ -61,7 +61,7 @@ data PushStrategy m r
   = PushStrategy
       { -- | Called when a path is already in the cache.
         onAlreadyPresent :: m r,
-        onAttempt :: RetryStatus -> m (),
+        onAttempt :: RetryStatus -> Int64 -> m (),
         on401 :: m r,
         onError :: ClientError -> m r,
         onDone :: m r,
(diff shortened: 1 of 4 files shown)


cachix/src/Cachix/Client/Push.hs
```diff
@@ -119,16 +119,19 @@ uploadStorePath ::
...
-uploadStorePath clientEnv _store cache cb storePath retrystatus = do
+uploadStorePath clientEnv store cache cb storePath retrystatus = do
   let (storeHash, storeSuffix) = splitStorePath $ toS storePath
       name = pushCacheName cache
-  onAttempt cb retrystatus
   narSizeRef <- liftIO $ newIORef 0
   fileSizeRef <- liftIO $ newIORef 0
   narHashRef <- liftIO $ newIORef ("" :: ByteString)
   fileHashRef <- liftIO $ newIORef ("" :: ByteString)
+  pathinfo <- liftIO $ Store.queryPathInfo store $ toS storePath
```
roberth: `queryPathInfo` does not tolerate subpaths or symlinks. I think it's taken care of elsewhere but you may want to make sure.

PR #270, opened 2020-01-27
narURL api

domenkozar:


cachix-api/src/Cachix/Api.hs
```diff
@@ -69,6 +69,12 @@ data BinaryCacheAPI route
             :> "nix-cache-info"
             :> Get '[XNixCacheInfo, JSON] NixCacheInfo,
         -- Hydra: src/lib/Hydra/View/NARInfo.pm
```
roberth: Doesn't the comment above belong to the code below the change?

cachix-api/src/Cachix/Api.hs
```diff
@@ -86,6 +86,12 @@ data BinaryCacheAPI route
             :> Summary "Given a list of store hashes, return a list of those that are missing"
             :> ReqBody '[JSON] [Text]
             :> Post '[JSON] [Text],
+        narURL
+          :: route
+               :- CachixAuth
+               :> "narurl"
+               :> Capture "nar" NarFileName
+               :> Get '[JSON] Text,
```
roberth: The `Text` here is probably a URL. Why is it special or when does a cachix client need to use this endpoint?

domenkozar: Client doesn't use this endpoint.

PR #283, opened 2020-02-28
Push jobs

domenkozar:


cachix/src/Cachix/Client/OptionsParser.hs
```diff
@@ -102,13 +103,22 @@ parserCachixCommand =
...
                 "The compression level for XZ compression.\
                 \ Take compressor *and* decompressor memory usage into account before using [7..9]!"
               <> showDefault
               <> value 2
           )
+        <*> option
+          auto
+          ( long "jobs"
+              <> short 'j'
+              <> help "Number of concurrent threads used for pushing store paths."
```
roberth:
```suggestion
              <> help "Number of threads used for pushing store paths."
```
Concurrency of threads is implied. The purpose here is paralellism if you want to mention that term, but it seems superfluous.

Not a big fan of "jobs", but I guess that's just "unavoidable" unix legacy in this case :(

PR #285, opened 2020-03-10
Fix #251: Assert nar hash before creating narinfo

### LINKED ISSUE #251: Bad archive: input doesn't look like a Nix archive
It seems that I managed to upload a broken `nar` to the cache with cachix (version 0.3.4): https://tenpureto.cachix.org/wrjd3z9x1gix6bxy0m9p4djzjypkrpxh.narinfo
Unlike some previous issues (#147), the file is not empty, just the [content](https://tenpureto.cachix.org/nar/bd11361f7dc8c701ddae2ea82853d63c07ad949051724f8d1fad30f8195c97a8.nar.xz) looks truncated, at least from the beginning.
The command line I used was
```
cachix push tenpureto -w &
nix-build --no-link -A ... | cachix push tenpureto
```
and it could have been also forcibly terminated it a couple of times. It runs on the CI, so I, unfortunately, have no chance to look at the nix store.

Will rebuilding and re-pushing to the cache work (I will re-build anyway, but it takes some time without the cache)? If not, can you please delete the broken archive?
If you want to investigate how it could happen, I will also be happy to help.

domenkozar:
Fixes uncommon but annoying bug of "bad nar archive" from Nix.

Never managed to reproduce. One theory is that the path disappears as
its deleted by GC or nix-store --delete.

### DIFF
--- a/cachix/src/Cachix/Client/Store.hs
+++ b/cachix/src/Cachix/Client/Store.hs
@@ -123,7 +123,7 @@ validPathInfoNarSize vpi =
         { (*$fptr-ptr:(refValidPathInfo* vpi))->narSize }
       |]
 
--- | The narHash.to_string field of a ValidPathInfo struct. Source: store-api.hh
+-- | Copy the narHash field of a ValidPathInfo struct. Source: store-api.hh
 validPathInfoNarHash :: ForeignPtr (Ref ValidPathInfo) -> IO ByteString
 validPathInfoNarHash vpi =
   unsafePackMallocCString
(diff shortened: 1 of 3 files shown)


cachix/src/Cachix/Client/Store.hs
```diff
@@ -122,6 +123,14 @@ validPathInfoNarSize vpi =
         { (*$fptr-ptr:(refValidPathInfo* vpi))->narSize }
       |]
 
+-- | The narHash.to_string field of a ValidPathInfo struct. Source: store-api.hh
```
roberth:
```suggestion
-- | Copy the narHash field of a ValidPathInfo struct. Source: store-api.hh
```

PR #286, opened 2020-03-11
Get deriver and references from Nix

domenkozar:


cachix/src/Cachix/Client/Store.hs
```diff
@@ -131,6 +133,27 @@ validPathInfoNarHash vpi =
...
+          return strdup((deriver == "" ? "unknown-deriver" : deriver->c_str()));
+        }
+      |]
+
+-- | References field of a ValidPathInfo struct. Source: store-api.hh
+validPathInfoReferences :: ForeignPtr (Ref ValidPathInfo) -> IO PathSet
+validPathInfoReferences vpi = do
+  ptr <-
+    [C.exp| const PathSet*
+            { &((*$fptr-ptr:(refValidPathInfo* vpi))->references) }
```
roberth: The `PathSet` wrapper assumes responsibility for `delete`ing its input, so we must copy it to avoid a double free. Something like:

```suggestion
            { new PathSet((*$fptr-ptr:(refValidPathInfo* vpi))->references) }
```

PR #298, opened 2020-05-01
api: migrate to nix-narinfo and SimpleNarInfo

sorki:
I've pushed it to Hackage as well few minutes ago, marking as draft until it lands in nixpkgs.

### DIFF
--- a/cachix-api/cachix-api.cabal
+++ b/cachix-api/cachix-api.cabal
@@ -39,6 +39,7 @@ library
   exposed-modules:
     Cachix.Api
     Cachix.Api.Error
+    Cachix.Api.NarInfo
     Cachix.Api.Signing
     Cachix.Api.Types
     Cachix.Types.BinaryCacheAuthenticated
@@ -69,6 +70,7 @@ library
     , jose
     , lens
     , memory
+    , nix-narinfo
     , protolude
     , resourcet
     , servant               >=0.14.1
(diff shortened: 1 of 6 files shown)


sorki:
Landed in master, is there a way to re-trigger actions?

domenkozar:
You'll need to update haskell.nix pin via `niv update haskell.nix`

sorki:
Looks like it's not yet in `haskell.nix`, we can add overlay or just wait since it's not that important.

domenkozar:
Thanks! It seems like the package is disabled in nixpkgs due to tests failing:

```
Test suite nix-narinfo-samples: RUNNING...

Parse
  roundtrips samples FAILED [1]

Failures:

  test/ParseSpec.hs:24:3:
  1) Parse roundtrips samples
       uncaught exception: IOException of type NoSuchThing
       ./test/samples/0: openFile: does not exist (No such file or directory)

  To rerun use: --match "/Parse/roundtrips samples/"

Randomized with seed 1178892351

Finished in 0.0002 seconds
1 example, 1 failure
```

sorki:
Oh, `extra-source-files` strikes again. Will fix, thanks!

sorki:
Had to bump to master at https://github.com/NixOS/nixpkgs/commit/5897105 since its not in unstable yet.

PR #304, opened 2020-05-25
Add --omit-deriver

roberth:
This adds an option to unset the deriver field. It's useful as an _extra_ line of defense for those who distribute binaries through cachix and have to deal with confidential source code.

### DIFF
--- a/cachix-api/test/Main.hs
+++ b/cachix-api/test/Main.hs
@@ -2,7 +2,6 @@ module Main where
 
 import Protolude
 import qualified Spec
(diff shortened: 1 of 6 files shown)


domenkozar:
For some reason CI won't even start. Diff looks good to me so I'm merging. Thanks!

domenkozar:
Ah, figured it out. `macos` is no longer a valid OS, but `macos-latest` is.

roberth:
Thanks!

PR #306, opened 2020-05-27
Fixed wrong suggestion for trusted users

Laobiz:
If you come across this error the suggestion will append "trusted-users = root $USER" with the double ticks into nix.conf. Somehow the escaping will be printed out with the string.

To avoid this we can use single ticks instead of double ticks.


domenkozar:
I think this was fixed by https://github.com/cachix/cachix/commit/1b8e7a6c1845391665d0d070423d866da7689a0b, but I'd need to verify.

domenkozar:
Improved the error message in cbc239d5954704ddcb0a7d552dfd86c41eb5b12d

Issue #314, opened 2020-06-19
Debugging binary cache configuration problems


domenkozar:
- In our project dir we set the `NIX_CONF_DIR` env variable to point towards a project configuration that sets the subsituters. However my user nix-configuration also contained a `trusted-public-keys` entry that overwrote the project-specific one.

roberth:
You can retrieve info using Nix's own logic using `Hercules.CNix.Settings.{getSubstituters,getTrustedPublicKeys}`.

PR #315, opened 2020-06-19
Add command to list caches

lazamar:
It seems useful to have a command to show what caches I'm using. 
I've wondered that before and had to look at the code to see the directory where the caches are stored.
This PR implements a simple command to show these.


domenkozar:
Note that this will only show caches that you have generated a key for. You'll see a more complete list on https://app.cachix.org

Are you interested into the caches you own or the caches that you have configured for Nix to use?

lazamar:
The caches Nix can use, so that I can know whether I've added all the caches I need/want.

domenkozar:
In that case it's more complicated, something I want to do as part of https://github.com/cachix/cachix/issues/314

domenkozar:
I'm closing this for now as if we do implement listing of caches, the correct implementation is much more involved.

PR #319, opened 2020-07-09
add temp CI job to test syspolicy impact

abathur:
Starting in Catalina, macOS runs a syspolicyd "assessment" that hits the network for each binary/script executable. It does cache these results, but Nix tends to introduce many "new" executables per build. (You can read more about this at https://github.com/NixOS/nix/issues/3789).

This PR adds a temporary, redundant macOS job with these assessments disabled. I'm hoping you can adopt it for a few weeks to help me collect more data on how this affects real projects.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,3 +18,20 @@ jobs:
         name: cachix
         signingKey: '${{ secrets.CACHIX_SIGNING_KEY }}'
     - run: nix-build ci.nix
+  macos_perf_test:
+    runs-on: macos-latest
+    steps:
+    - name: Disable syspolicy assessments
+      run: |
+        spctl --status
+        sudo spctl --master-disable
+    - uses: actions/checkout@v2.3.1
+    - uses: cachix/install-nix-action@v10
+      with:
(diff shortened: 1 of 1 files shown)


domenkozar:
It looks like it finally get performance on pair with linux. 

How about we add that to https://github.com/cachix/install-nix-action?

abathur:
Well, I'm hoping that collecting this data will demonstrate savings sufficiently large to justify incorporating this workaround into Nix itself (or demonstrate that my project is an outlier, and that most people aren't seeing enough impact to bother. I've sent a similar PR to 11 total projects (4 have merged the PR so far), and I'm also tracking one of my own.

I would prefer no change to `install-nix-action` _for now_, so that I don't have to worry about any of these projects updating to the new action and messing up the run-duration data I'm collecting. 

If there's resistance to addressing this in Nix, I completely agree with doing it by default in `install-nix-action`.

I've been running the job for 2 weeks on my project and I've found the durations are very noisy run-to-run/job-to-job (I've seen a fair number of builds where the disabled build is slower), so I'm hoping to get enough samples to be more confident about the size of the effect, and how much it varies project to project.

PR #327, opened 2020-08-19
Add flake.nix

matthewbauer:
This is mainly so `nix profile install github:cachix/cachix` works correctly with cachix.


flake.nix
```diff
@@ -0,0 +1,20 @@
...
+  outputs = { self }: let
+    systems = [ "x86_64-linux" "i686-linux" "x86_64-darwin" "aarch64-linux" ];
+    forAllSystems = f: builtins.listToAttrs (map (name: { inherit name; value = f name; }) systems);
+  in {
+    packages = forAllSystems (system: {
+      cachix = let
+        pkgs = import ./nix {
+          inherit system;
+        };
+        cachix-api = pkgs.haskellPackages.callCabal2nix "cachix-api" (pkgs.gitignoreSource ./cachix-api) {};
```
domenkozar, line 13: Is there any way to deduplicate this against default.nix? I'll surely forget to update it the next time.

On top of that CI should also build the flake bit I guess, so maybe we could use flake compat?

newkozlukov: If I understand it right, with flakes there isn't much need in `niv` and it should be possible move all external dependencies into `inputs` attribute of `flake.nix`. The specific revisions would be pinned in `flake.lock`

domenkozar:
@matthewbauer do you have an opinion how to deduplicate stuff?

matthewbauer:
Sorry for the delay!

I updated it to import cachix from default.nix.

PR #328, opened 2020-08-31
Revert "add temp CI job to test syspolicy impact"

abathur:
Reverts #319

Thanks for your help testing this out. The results were a bit of a bust with respect to any short-term hope of speeding up macOS builds, but they did help me tease out the fact that the approach I used here doesn't actually disable the whole assessment--it seems to just keep the result of the assessment from being used.

With the approach used here ruled out, the only way I'm aware of to fully disable the assessments and reclaim the associated time requires a GUI to set up. :/


domenkozar:
Thanks for your work :)

PR #351, opened 2020-12-09
Fix instructions for adding user to trustedUsers

3noch:
I don't know if this is correct for all versions of Nixpkgs but it's what I had to do on 20.09.


domenkozar:
Thanks!

Issue #339, opened 2020-09-27
`push -w` on remote builder leads to system crash (out of memory)


DavHau:
Hey, thanks for cachix!

I'm currently integrating it into my remote builder infrastructure.
To push my build results directly to cachix, I run `cachix -w --compression-level 0 {cache}` on the remote builder.

Problem:
As soon as I start a bigger build, cachix starts endless amount of processes, depleting the systems memory after a short time resulting in a total crash of the remote builder.

Not using the `cachix push -w` during the build and instead running `cachix push` after the build finishes, works fine.

I'm using cachix 0.3.8

domenkozar:
Could you make a screenshot of processes/ resource usage? Some more information would help to better understand what's going on.

domenkozar:
@DavHau I'm happy to help with this, but need to be able to reproduce this. Maybe we could do a shared video session?

DavHau:
Sorry for the delay. Here a screenshot of htop while provoking the previously explained behaviour:
![image](https://user-images.githubusercontent.com/42246742/94938784-82a3b000-04fb-11eb-88af-df1df8733c11.png)
It just goes on and on like this, if i scroll down.


And shortly after, the machine is crashed:
![image](https://user-images.githubusercontent.com/42246742/94938985-b7b00280-04fb-11eb-9f95-add856b5ae9e.png)
Cachix is running inside a docker container.
I might be able to prepare a docker container for you which triggers the error, if that helps.

In general I'm working on a remote builder with fast cachix integration.
To already start uploading while a build is still in progress, my Idea was to run `cachix push -w` whenever a build starts. A finishing build would kill the `cachix push -w` process, and then push the remaining artifacts via a subsequent `cachix push`.

But even if no crash occurs, my strategy doesn't seem to work. Some files seem to be missing on cachix afterwards. Not exactly sure why. Maybe the cachix backend signals the second pushing process, that some files already exist, even they had only been partly uploaded by the previous process (which was killed)?

sure, we can make a video session if you like.
FYI, It's not a blocking issue on my side. Instead of pushing to cachix while building, I'm OK with pushing after the build, which works fine. This would just be an optimization for me.

domenkozar:
This would be solved with #53 at least for better ergonomics.

Just to be clear, what does `build` mean in this context, `nix-build` invocation?

If you can prepare a docker image that would be ideal to observe. Or a video session would work :)

[...]

Issue #343, opened 2020-11-03
support a more selective watch mode via a queue+socket+post-build-hook


colemickens:
Idea:

I really want to capture build outputs as they're produced, but I don't really like that `cachix watch` picks up things from unrelated builds.

It seems like this be fixed by utilizing a post build script. However, I don't like configuring it on the daemon because it's one extra step to do on remote builders, it would catch "all" builds again, and I try to keep my "buildargs" portable so that I always get cache hits, etc (this might be alleviated once flakes respects custom nix conf).

But, it seems like we would need to "queue" the signing+pushing still and do it out-of-band from the builds. That's where `cachix` can play a role:

1. Start cachix, listen on a socket, sign+push paths when told about them over the socket
2. Let the user use the post-build-hook to send paths to that socket on success nix build

```bash
# untested, example (can't remember if I need to force it to not interpolate OUT_PATHS yet)

cat<<EOF >/tmp/script.sh
echo "${OUT_PATHS[@]}"  | cachix --queue-socket /tmp/cachix.sock"
EOF

cachix --watch-socket /tmp/cachix.sock
nix build '.#hello' -L --option post-build-hook "/tmp/script.sh"
```

domenkozar:
Is post-build-hook executed on the client or daemon side?

domenkozar:
(you probably mean cachix where you say cargo?)

colemickens:
yes, `s/cargo/cachix/g`, updated the original post.

I'm not sure if it's executed client side or daemon side, but if I stash `whoami` in the post-build-hook into a file and then check it, it says `root`...

domenkozar:
Yeah I agree with this and I plan to experiment how feasible it is. I've laid down a plan in #53 and this is on the horizon soon!

domenkozar:
Here's the plan how this will work:

Using async to start two concurrent threads. The first thread is doing stuff and feeding things to a channel. The second one is consuming from channel and doing things based on those message. To handle done case: `data Message = Done | Do Thing` would be the message passed, so second thread would exit on `Done` case.

I have then yet to figure out how to handle CTRL-C but this makes the happy case smooth.

domenkozar:
Going to start next week.

domenkozar:
Queue implemented in https://github.com/cachix/cachix/pull/353, what's left then is the interface for post-build-hook and sockets.

@colemickens are you sure that post-build-hook works when specified on the client?

colemickens:
[...]

PR #312, opened 2020-06-10
Stream Nar natively

domenkozar:
Needs to compile first. Then needs testing.

Depends on https://github.com/haskell-nix/hnix-store/pull/61

PR #353, opened 2021-01-05
push watch: use queue when pushing, wait on queue to empty upon ctrl-c

### LINKED ISSUE #53: Add signal handling to the daemon
In order to avoid having to call `cachix push` [twice](51), we can do the following:

* Add signal handling to the daemon mode
* Have a signal that, when received, finishes pushing and then exits
* Run the daemon in a job in parallel with `nix-build`
* Send that signal after `nix-build`

domenkozar:
This also contains some breaking changes:

- `Push` module now operates with `PushCache m r` that simplifies how many arguments need to be passed around
- watch mode doesn't upload `drv` files anymore

Fixes #53 #339

Example:

```
Watching /nix/store for new paths ...
compressing and pushing /nix/store/g6si4dzzxvv0vibrc1sa1qdirgxv9v1i-1609944078-1 (976.67 KiB)
compressing and pushing /nix/store/2pf7xljj84z5f0880bsgkbipndah39zp-empty-10-1MB (680.00 B)
compressing and pushing /nix/store/243apdjx4ddwk45kijmlc99kdnsy5d24-1609944078-7 (976.67 KiB)
compressing and pushing /nix/store/6x43igsqvg7g6j5rnpa0g4m6rbqm5nbk-1609944078-3 (976.67 KiB)
compressing and pushing /nix/store/baabkwgbrmjz9m672x3x57pdjp0ah9x5-1609944078-9 (976.67 KiB)
^CStopping producer of store paths and waiting for queue to empty ...
Waiting to finish: 5 pushing, 6 in queue
compressing and pushing /nix/store/c3q7dnd9zg32azz1jglyzdi8yi5yzi5v-1609944078-5 (976.67 KiB)
Waiting to finish: 5 pushing, 5 in queue
compressing and pushing /nix/store/h2x8ibnwvswnv16wry71fn6vajljqdqg-1609944078-10 (976.67 KiB)
Waiting to finish: 5 pushing, 4 in queue
compressing and pushing /nix/store/k859flmfrnppmf4r4w0x20sp7k0cn8lg-1609944078-6 (976.67 KiB)
Waiting to finish: 5 pushing, 3 in queue
compressing and pushing /nix/store/plnp0l8y4h9scl7zzvjr4i7h8fcyj1yb-1609944078-8 (976.67 KiB)
Waiting to finish: 5 pushing, 2 in queue
compressing and pushing /nix/store/pzakihpgprjvfs92z1hpdr4mgcsnhjza-1609944078-2 (976.67 KiB)
compressing and pushing /nix/store/yvw3may5yphfxfpi1dqfb5xx7w8zvd2z-1609944078-4 (976.67 KiB)
Waiting to finish: 5 pushing, 0 in queue
Waiting to finish: 5 pushing, 0 in queue
Waiting to finish: 4 pushing, 0 in queue
Waiting to finish: 1 pushing, 0 in queue
Waiting to finish: 1 pushing, 0 in queue
Waiting to finish: 1 pushing, 0 in queue
Waiting to finish: 1 pushing, 0 in queue
Waiting to finish: 1 pushing, 0 in queue
Waiting to finish: 1 pushing, 0 in queue
Waiting to finish: 1 pushing, 0 in queue
All done, exiting.
```

### DIFF
--- a/cachix/src/Cachix/Client/PushQueue.hs
+++ b/cachix/src/Cachix/Client/PushQueue.hs
@@ -39,36 +39,36 @@ data QueryWorkerState
         lock :: Lock.Lock
       }
 
-worker :: Push.PushCache IO () -> PushWorkerState -> IO ()
-worker pushCache workerState = forever $ do
+worker :: Push.PushParams IO () -> PushWorkerState -> IO ()
+worker pushParams workerState = forever $ do
   storePath <- atomically $ TBQueue.readTBQueue $ pushQueue workerState
   bracket_ (inProgresModify (+ 1)) (inProgresModify (\x -> x - 1))
     $ retryAll
-    $ Push.uploadStorePath pushCache storePath
+    $ Push.uploadStorePath pushParams storePath
   where
     inProgresModify f =
       atomically $ modifyTVar' (inProgress workerState) f
 
 -- NOTE: producer is responsible for signaling SIGINT upon termination
 -- NOTE: producer should return an `IO ()` that should be a blocking operation for terminating it
-startWorkers :: Int -> (Queue -> IO (IO ())) -> Push.PushCache IO () -> IO ()
-startWorkers numWorkers mkProducer pushCache = do
+startWorkers :: Int -> (Queue -> IO (IO ())) -> Push.PushParams IO () -> IO ()
+startWorkers numWorkers mkProducer pushParams = do
   -- start query worker
   (newQueryQueue, newPushQueue, newLock) <-
     atomically $
       (,,) <$> TBQueue.newTBQueue 10000 <*> TBQueue.newTBQueue 10000 <*> Lock.new
   let queryWorkerState = QueryWorkerState newQueryQueue S.empty newLock
-  queryWorker <- async $ queryLoop queryWorkerState newPushQueue pushCache
+  queryWorker <- async $ queryLoop queryWorkerState newPushQueue pushParams
   -- start push workers
   stopProducerCallback <- mkProducer newQueryQueue
   progress <- newTVarIO 0
   let pushWorkerState = PushWorkerState newPushQueue progress
-  pushWorker <- async $ replicateConcurrently_ numWorkers $ worker pushCache pushWorkerState
+  pushWorker <- async $ replicateConcurrently_ numWorkers $ worker pushParams pushWorkerState
   void $ Signals.installHandler Signals.sigINT (Signals.CatchOnce (exitOnceQueueIsEmpty stopProducerCallback pushWorker queryWorker queryWorkerState pushWorkerState)) Nothing
   waitEither_ pushWorker queryWorker
 
-queryLoop :: QueryWorkerState -> Queue -> Push.PushCache IO () -> IO ()
-queryLoop workerState pushqueue pushCache = do
+queryLoop :: QueryWorkerState -> Queue -> Push.PushParams IO () -> IO ()
+queryLoop workerState pushqueue pushParams = do
   -- this blocks until item is available and doesn't remove it from the queue
   _ <- atomically $ TBQueue.peekTBQueue (queryQueue workerState)
   (missingStorePathsSet, alreadyQueuedSet) <- Lock.with (lock workerState) $ do
@@ -79,12 +79,12 @@ queryLoop workerState pushqueue pushCache = do
       if isEmpty
         then return S.empty
         else return $ alreadyQueued workerState
-    missingStorePaths <- Push.getClosure pushCache storePaths
+    missingStorePaths <- Push.getMissingPathsForClosure pushParams storePaths
     let missingStorePathsSet = S.fromList missingStorePaths
         uncachedMissingStorePaths = S.difference missingStorePathsSet alreadyQueuedSet
     atomically $ for_ uncachedMissingStorePaths $ TBQueue.writeTBQueue pushqueue
     return (missingStorePathsSet, alreadyQueuedSet)
-  queryLoop (workerState {alreadyQueued = S.union missingStorePathsSet alreadyQueuedSet}) pushqueue pushCache
+  queryLoop (workerState {alreadyQueued = S.union missingStorePathsSet alreadyQueuedSet}) pushqueue pushParams
 
 exitOnceQueueIsEmpty :: IO () -> Async () -> Async () -> QueryWorkerState -> PushWorkerState -> IO ()
 exitOnceQueueIsEmpty stopProducerCallback pushWorker queryWorker queryWorkerState pushWorkerState = do
--- a/cachix/src/Cachix/Client/PushQueue.hs
+++ b/cachix/src/Cachix/Client/PushQueue.hs
@@ -1,0 +1,73 @@
+{- Implements a queue with the following properties:
+
+- waits for queue to be fully pushed when exiting using ctrl-c (SIGINT)
+- allows stopping the producer
+- avoid duplicate pushing of the same store paths
+
+To safetly exit on demand, signal SIGINT.
+-}
+module Cachix.Client.PushQueue
+  ( pushQueue,
+    Queue,
+  )
+where
+
+import qualified Cachix.Client.Push as Push
+import Control.Concurrent.Async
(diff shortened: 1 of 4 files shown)


cachix/src/Cachix/Client/PushQueue.hs
```diff
@@ -0,0 +1,73 @@
...
+  = WorkerState
+      { inQueue :: Queue,
+        inProgress :: TVar (S.Set StorePath)
+      }
+
+worker :: Push.PushCache IO () -> WorkerState -> IO ()
+worker pushCache workerState = forever $ do
+  storePath <- atomically $ TBQueue.readTBQueue $ inQueue workerState
+  inprogress <- readTVarIO $ inProgress workerState
+  -- TODO: there's a small window between checking membership and inserting new path
```
roberth, line 37: Your first `atomically` can take from the queue, update `inProgress` and return `Maybe StorePath`.
Then `for` the maybe, you do the work and at then end update `inProgress` in the second transaction.

cachix/src/Cachix/Client/PushQueue.hs
```diff
@@ -0,0 +1,73 @@
...
+  unless (S.member storePath inprogress) $ bracket_ (start storePath) (stop storePath) $
+    Push.pushSingleStorePath pushCache storePath
+  where
+    start storePath =
+      atomically $ modifyTVar' (inProgress workerState) (S.insert storePath)
+    stop storePath =
+      atomically $ modifyTVar' (inProgress workerState) (S.delete storePath)
+
+-- NOTE: producer is responsible for signaling SIGINT upon termination
+pushQueue :: Int -> (Queue -> IO (IO ())) -> Push.PushCache IO () -> IO ()
```
roberth, line 47: Perhaps `withNewPushWorker`

cachix/src/Cachix/Client/PushQueue.hs
```diff
@@ -0,0 +1,73 @@
...
+  stopProducerCallback
+  go
+  where
+    go = do
+      queueLength <- atomically $ TBQueue.lengthTBQueue $ inQueue workerState
+      inprogress <- readTVarIO $ inProgress workerState
+      if queueLength == 0 && S.null inprogress
+        then do
+          putText "All done, exiting."
+          uninterruptibleCancel consumer
```
roberth, line 69: Potential race condition here. It's nicer to put an "EOF" on the queue, so the threads can exit by themselves and you just wait for the consumer async to complete here.

domenkozar: What is the race condition that can happen?

roberth: I guess this is more of a general advice than specific to cachix.

 - thread a creates thread b
 - thread b adds to queue
 - thread a terminates the worker
 - thread a assumes everything is done
 - thread b adds to queue

It probably doesn't apply here.

domenkozar: On line 60 it terminates the producer, I should note that it should be a blocking action.

I tried the approach with EOF but it makes it more complicated to report number of items in queue.

domenkozar:
Also need to implement batching for narinfos.

domenkozar:
@roberth still need to figure out how to wait both queues to finish, otherwise query queue is there.

cachix/src/Cachix/Client/PushQueue.hs
```diff
@@ -0,0 +1,102 @@
...
+  consumer <- async $ replicateConcurrently_ numWorkers $ worker pushCache workerState
+  void $ Signals.installHandler Signals.sigINT (Signals.CatchOnce (exitOnceQueueIsEmpty stopProducerCallback consumer workerState)) Nothing
+  waitEither_ consumer queryWorker
+
+queryLoop :: QueryWorkerState -> Queue -> Push.PushCache IO () -> IO ()
+queryLoop workerState pushQueue pushCache = do
+  storePaths <- atomically $ do
+    -- this blocks until item is available and doesn't remove it from the queue
+    _ <- TBQueue.peekTBQueue (queryQueue workerState)
+    TBQueue.flushTBQueue (queryQueue workerState)
```
roberth, line 72: Equivalent to flush + `guard` not empty, fwiw

roberth:
> @roberth still need to figure out how to wait both queues to finish, otherwise query queue is there.

You could probably introduce a `TVar Bool` to indicate whether you have in-flight queries and check that one as well in the same `atomically` as where you check that the worker queue is empty. And you should check the `queryQueue` itself there as well. That should do.

If you want a nicer design, you could consider tracking the status of the store paths themselves in `TVar`s that accompany the `StorePath`s. That way you can decouple the termination logic from the queues.

cachix/src/Cachix/Client/PushQueue.hs
domenkozar: Flush says it never retries, so it would then need handle the loop

domenkozar:
@roberth alright this is ready - there's one problem though, sigint is still caught in the subprocess `nix-store --dump` so maybe time to revive #312

cachix/src/Cachix/Client/PushQueue.hs
roberth: The `guard` would retry it.

roberth:
Sounds like `delegate_ctlc` in `System.Process`. http://hackage.haskell.org/package/process-1.6.10.0/docs/System-Process.html#g:4

domenkozar:
That one doesn't help, but I found out that `create_group` makes it ignore signals. Probably not the best idea in case those processes hang?

domenkozar:
Alright it should still work since removing stdin makes them crash. Not the best, but it works.

roberth:
> That one doesn't help, but I found out that `create_group` makes it ignore signals. Probably not the best idea in case those processes hang?

If you want to protect against hanging you'd need a `timeout` and an extra exception handler because the one in `createProcess` only sends a SIGTERM iirc. I don't think you want to be paranoid like that though.

roberth:
Although the `timeout` isn't necessary when your Ctrl+C causes the exception.

domenkozar:
I've tested SIGINT and SIGTERM and it behaves as expected now.

domenkozar:
Did more testing, it works! @roberth any objections for merging?

roberth (commented):
Looks alright with the small naming improvements.

I might still prefer a custom solution for hercules-ci-agent, because in a multi-process situation I want to know when a path is done without shutting down the worker. So I'll probably use the low-level `uploadStorePath`/`pushSingleStorePath` and call the `narInfoBulk` endpoint myself as well.
I'd like to be able to reuse `PushQueue` but I feel like that's unlikely to be a good fit with support for other caches and the multi-process nature of the agent. Also Hercules CI already uploads store paths immediately so the immediate benefit is small. (more efficient bulk queries)

To summarize, I think this is a good solution for the `cachix` CLI and I can still use it on the agent. So no objections :)

domenkozar:
> I might still prefer a custom solution for hercules-ci-agent, because in a multi-process situation I want to know when a path is done without shutting down the worker. So I'll probably use the low-level uploadStorePath/pushSingleStorePath and call the narInfoBulk endpoint myself as well.

If you come up with a way to generalize the current solution that seems acceptable to me. I also want to support #343 that will require changes to work in a daemon mode.

roberth:
> > I might still prefer a custom solution for hercules-ci-agent, because in a multi-process situation I want to know when a path is done without shutting down the worker. So I'll probably use the low-level uploadStorePath/pushSingleStorePath and call the narInfoBulk endpoint myself as well.
> 
> If you come up with a way to generalize the current solution that seems acceptable to me. I also want to support #343 that will require changes to work in a daemon mode.

If the caller can provide a `(StorePath, TVar Bool)` instead of just a storepath and the code sets it to true when done, that should go a long way, but I'm not sure if that's worthwhile for hercules-ci-agent because it needs to generalize to other cache types besides cachix. I suppose that might be useful in for your #343 too. Idk

domenkozar:
Other caches don't have a good way to query for closure in efficient way, but lets leave generalizations for when that's needed :)

PR #361, opened 2021-03-05
Profiling and cnix store

roberth:
This PR provides a couple of improvements:
 - Update to GHC 8.10.4
 - Move the custom libstore binding into an external library that I maintain `hercules-ci-cnix-store`
 - Allow `cachix` to be built with profiling enabled (revert 712432e35a340acb23f933e52409c65856d37364 if you want to disable it by default)
 - Allow `ci.nix` to perform a build on the three main systems simultaneously for use in Hercules CI
 - Makes the CI complete again. Includes and closes #359


roberth:
Cool. I'll rebase.

The `hercules-ci-` prefix is there because it's not an official binding. The vision is simple: just provide a binding to libstore. Anything that's hercules-specific doesn't belong in it. For example, hercules-ci-agent keeps a couple of its own inline-c-cpp functions, so you don't have to worry about library bloat.

PR #368, opened 2021-04-04
cachix authtoken now reads from stdin if no token is provided on the command line

### LINKED ISSUE #363: Support authtoken on stdin
The CI platform I use (builds.sr.ht) has `set -x` on (which is useful!).
However, this results in a command like `cachix authtoken $(< ~/.cachix-auth-token)` leaking the auth token in the logs, since `set -x` logs after expansion of variables, command substitutions, etc.

Would it be reasonable to read the auth token from stdin if none is provided to the CLI, or if `--stdin` is passed or something?
(I'll submit a PR if so.)

remexre:
Closes #363.

Should I change the docs to recommend this as well? Passing secrets as command line arguments is, in general, insecure, since other users can snoop on them via `/proc/*/cmdline`.

### DIFF
--- a/cachix/src/Cachix/Client/Commands.hs
+++ b/cachix/src/Cachix/Client/Commands.hs
@@ -67,7 +67,7 @@ authtoken env (Just token) = do
   writeConfig (configPath (cachixoptions env)) $ case config env of
     Just cfg -> Config.setAuthToken cfg $ Token (toS token)
     Nothing -> mkConfig token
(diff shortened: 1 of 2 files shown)


domenkozar:
Changing `cachix authtoken --help` to mention stdin fallback would be great.

On second thought, `--stdin` would be preferred as just `cachix auttoken` will one day do oauth cli flow.

PR #383, opened 2021-08-09
use systemd to notify shutdown for watch commands

domenkozar:
Allows clean shutdown when used with systemd. cc @mic92 for testing

### DIFF
--- a/cachix/src/Cachix/Client/PushQueue.hs
+++ b/cachix/src/Cachix/Client/PushQueue.hs
@@ -61,7 +61,9 @@ startWorkers numWorkers mkProducer pushParams = do
   progress <- newTVarIO 0
   let pushWorkerState = PushWorkerState newPushQueue progress
   pushWorker <- async $ replicateConcurrently_ numWorkers $ worker pushParams pushWorkerState
-  void $ Signals.installHandler Signals.sigINT (Signals.CatchOnce (exitOnceQueueIsEmpty stopProducerCallback pushWorker queryWorker queryWorkerState pushWorkerState)) Nothing
+  let signalset = Signals.CatchOnce (exitOnceQueueIsEmpty stopProducerCallback pushWorker queryWorker queryWorkerState pushWorkerState)
+  void $ Signals.installHandler Signals.sigINT signalset Nothing
+  void $ Signals.installHandler Signals.sigTERM signalset Nothing
   (_, eitherException) <- waitAnyCatchCancel [pushWorker, queryWorker]
   case eitherException of
     Left exc | fromException exc == Just StopWorker -> return ()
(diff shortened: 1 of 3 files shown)


cachix/src/Cachix/Client/PushQueue.hs
```diff
@@ -95,6 +96,7 @@ queryLoop workerState pushqueue pushParams = do
 exitOnceQueueIsEmpty :: IO () -> Async () -> Async () -> QueryWorkerState -> PushWorkerState -> IO ()
 exitOnceQueueIsEmpty stopProducerCallback pushWorker queryWorker queryWorkerState pushWorkerState = do
   putText "Stopped watching /nix/store and waiting for queue to empty ..."
+  Systemd.notifyStopping
```
Mic92, line 99: Is this the haskell counterpart for?

```
           If a service of Type=notify sends "EXTEND_TIMEOUT_USEC=...", this
           may cause the stop time to be extended beyond TimeoutStopSec=. The
           first receipt of this message must occur before TimeoutStopSec= is
           exceeded, and once the stop time has extended beyond
           TimeoutStopSec=, the service manager will allow the service to
           continue to stop, provided the service repeats
           "EXTEND_TIMEOUT_USEC=..."  within the interval specified, or
           terminates itself (see sd_notify(3)).
```

This is from `systemd.service(5)` manpage.

Mic92: Seems to be not implemented yet but https://hackage.haskell.org/package/systemd-2.3.0/docs/src/System.Systemd.Daemon.html#notifyStatus
allows to implement that. My haskell foo is not strong enough to come up with more concrete suggestions.

Mic92: But reading further below looks like this function would be the right place to extend the timeout.

domenkozar:
@Mic92 PTAL :)

cachix/src/Cachix/Client/PushQueue.hs
```diff
@@ -112,6 +116,8 @@ exitOnceQueueIsEmpty stopProducerCallback pushWorker queryWorker queryWorkerStat
           cancelWith queryWorker StopWorker
           cancelWith pushWorker StopWorker
         else do
+          -- extend shutdown for another 90s
+          Systemd.notify False $ "EXTEND_TIMEOUT_USEC=" <> show (90 * 1000 * 1000)
```
Mic92, line 120: I was not sure how systemd would interpret this timeout but what it does is according to the source:
`current_time + EXTEND_TIMEOUT_USEC`. So this implementation seems correct: if there is no cachix update for 90s
than the service is killed.

domenkozar: That was my interpretation from `the service manager will allow the service to
           continue to stop, provided the service repeats
           "EXTEND_TIMEOUT_USEC=..."  within the interval specified, or
           terminates itself (see sd_notify(3)).`

domenkozar: @Mic92 if this looks good to you I'll merge it and you can give it a try :)

Mic92: It does.

PR #384, opened 2021-08-11
Filter out invalid paths when pushing

domenkozar:
Fixes https://github.com/cachix/cachix-action/issues/83

### DIFF
--- a/cachix/src/Cachix/Client/CNix.hs
+++ b/cachix/src/Cachix/Client/CNix.hs
@@ -1,0 +1,22 @@
+module Cachix.Client.CNix where
+
+import qualified Data.Text as T
+import qualified Language.C.Inline.Cpp.Exceptions as C
+import Protolude
+import System.Console.Pretty (Color (..), color)
+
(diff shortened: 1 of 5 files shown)


domenkozar:
I'm not yet happy about this, as users won't notice that invalid paths aren't pushed.

Ideally, it would push all the rest of the paths and error out.

cachix/src/Cachix/Client/CNix.hs
```diff
@@ -0,0 +1,22 @@
...
+      then Just $ toS msg
+      else Nothing
+predicateInvalidPath _ = Nothing
+
+handleInvalidPath :: IO (Maybe a) -> IO (Maybe a)
+handleInvalidPath = handleJust predicateInvalidPath handle_
+  where
+    handle_ :: Text -> IO (Maybe a)
+    handle_ msg = do
+      putStrLn $ color Red $ "Warning: " <> msg
```
domenkozar, line 21: stderr, yellow color

PR #398, opened 2021-12-21
README: instruct how to install using the new interface

domenkozar:


pingiun (approved):
This is nice. I hope you can also have some way to install the latest release instead of master. Using a branch named `release` would enable `nix profile install github:cachix/cachix/release`

domenkozar:
Does it work with tags? `nix profile install github:cachix/cachix/v0.6.1`

pingiun:
> Does it work with tags? `nix profile install github:cachix/cachix/v0.6.1`

Yes that works but it will not be able to update to the latest release like the other installation method with `nix-env` does. Because if you run `nix profile upgrade` it will still pull the specified tag and see if it updated

domenkozar:
I could use something like https://github.com/marketplace/actions/latest-tag to maintain that automatically

pingiun:
Seems nice!

PR #402, opened 2022-02-04
Update NixOS installation to structural settings

jtojnar:
https://github.com/NixOS/nixpkgs/pull/139075


jtojnar:
Prevents warnings on nixos-unstable but it is not really compatible with stable NixOS.

domenkozar:
In that case I'd rather see warnings than generate something that doesn't work on stable channels.

roberth:
`nix.settings` shouldn't warn yet. I've opened https://github.com/NixOS/nixpkgs/pull/163451 to address this.

PR #416, opened 2022-07-07
Cachix deploy handle restarts

### LINKED ISSUE #411: Cachix Deploy: any deployment that restarts the agent on macOS disconnects & requires manual intervention
Any deployment to macOS that happens to restart the Cachix Deploy agent causes the agent to sign off and never reconnect, so I have to log in to each Mac and deploy locally (manually) to get the deployment to work and the agent to reconnect. When the agent reconnects, the deployment shows up in the UI as "Canceled", but that's just a minor issue compared to the hassle of needing to deploy manually.

It's possible we're doing something wrong in our config, but there's not much to configure, so it's not clear what it would be.

domenkozar:
Fixes #411 by having deployments a separate process

### DIFF
--- a/cachix/cachix-deployment/Main.hs
+++ b/cachix/cachix-deployment/Main.hs
@@ -30,8 +30,6 @@ import Protolude.Conv
 import System.IO (BufferMode (..), hSetBuffering)
 import qualified Wuss
 
--- TODO: https://superuser.com/questions/1333069/how-to-track-all-child-processes-spawned-by-a-systemctl-service
-
 main :: IO ()
 main = do
   setLocaleEncoding utf8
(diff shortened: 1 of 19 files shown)


domenkozar:
```
$ /nix/store/0g0s728s9znq3jifjfn4qvw9m3j28dy5-nixos-system-cherimoya-22.05pre-git/bin/switch-to-configuration switch
leaving systemd-boot 250.4 in place (250.4 is not newer)
stopping the following units: cachix-agent.service
activating the configuration...
setting up /etc...
reloading user units for domen...
setting up tmpfiles
starting the following units: cachix-agent.service
the following new units were started: thermald.service
Successfully activated the deployment.
```

Success!

PR #431, opened 2022-08-11
Remove the workaround for flushing logs when deploying

sandydoo:
The workaround was a delay, added at the end of the deployment activation, to give the logger some time to catch up with upstreaming the logs. This became necessary because the logging and activation threads were being run concurrently with `race`. As soon as the activation was complete, the logging thread would be immediately cancelled.

The fix is to run the threads `concurrently` and use a closable channel (TMQueue) to control the lifecycle of the logger, instead of relying on asynchronous exceptions.

The goodbye message for the deployment socket has also been moved out of the logger.


domenkozar:
Nice 🚀

PR #438, opened 2022-08-23
Use file locks to limit concurrent profile deployments

sandydoo:
This handles the edge case where the agent launches multiple concurrent deployments.

The design is as follows:
1. The agent launches the deployer binary as usual.
2. The deployer tries to acquire a lock on a lock file, stored in either the user cache directory or in `/var/run` and named after the target profile.
3. Either:
    - The lock is acquired and the deployment continues.
    - The lock is busy. The deployer exits immediately. The backend reschedules the deployment.

To avoid race conditions, the lock files are not removed.
If the owner of the lock crashes, the lock file is unlocked.

### Update

To avoid creating root-owned files inside user directories, we check that the home directory is owned by the current user. `withTryLock` should ideally sanity check this as well in case we remove the check from the agent.

Here’s what the error looks like on macOS:
```sh
❯ sudo result/bin/cachix deploy agent cachix-test-darwin test
[2022-08-31 16:03:07][agent.agent][Error][asdf.local][PID 8390][ThreadId 7][cachix-0.9.0-5e0pDsosASb1tBWytm0Ma9:Cachix.Deploy.Websocket src/Cachix/Deploy/Websocket.hs:47:16] The current user (root) does not own the home directory (/Users/Sander)
Try running the agent with `sudo -H`.
```

### DIFF
--- a/cachix/cachix-deployment/Main.hs
+++ b/cachix/cachix-deployment/Main.hs
@@ -45,7 +45,7 @@ main = do
   input <- escalateAs (FatalError . toS) . Aeson.eitherDecode . toS =<< getContents
   let profile = toS . CachixWebsocket.profile . Input.websocketOptions $ input
   void $
-    Lock.withTryLock profile $
+    Lock.withTryLock ("/nix/var/nix/profiles/" <> profile) $
       CachixWebsocket.runForever (CachixWebsocket.websocketOptions input) (handleMessage input)
 
 handleMessage :: CachixWebsocket.Input -> ByteString -> (K.KatipContextT IO () -> IO ()) -> WS.Connection -> CachixWebsocket.AgentState -> ByteString -> K.KatipContextT IO ()
--- a/cachix/cachix-deployment/Main.hs
+++ b/cachix/cachix-deployment/Main.hs
@@ -11,6 +11,7 @@ import Cachix.Client.Retry
 import qualified Cachix.Deploy.Activate as Activate
 import qualified Cachix.Deploy.Lock as Lock
 import qualified Cachix.Deploy.Websocket as CachixWebsocket
+import qualified Cachix.Deploy.Websocket as Input
 import Conduit ((.|))
 import qualified Control.Concurrent.Async as Async
 import qualified Control.Concurrent.STM.TMQueue as TMQueue
@@ -40,10 +41,11 @@ main = do
   setLocaleEncoding utf8
   hSetBuffering stdout LineBuffering
   hSetBuffering stderr LineBuffering
+
   input <- escalateAs (FatalError . toS) . Aeson.eitherDecode . toS =<< getContents
-  let store = toS . WSS.storePath . CachixWebsocket.deploymentDetails $ input
+  let profile = toS . CachixWebsocket.profile . Input.websocketOptions $ input
   void $
-    Lock.withTryLock store $
+    Lock.withTryLock profile $
       CachixWebsocket.runForever (CachixWebsocket.websocketOptions input) (handleMessage input)
 
 handleMessage :: CachixWebsocket.Input -> ByteString -> (K.KatipContextT IO () -> IO ()) -> WS.Connection -> CachixWebsocket.AgentState -> ByteString -> K.KatipContextT IO ()
--- a/cachix/cachix-deployment/Main.hs
+++ b/cachix/cachix-deployment/Main.hs
@@ -31,6 +31,10 @@ import Protolude.Conv
 import System.IO (BufferMode (..), hSetBuffering)
 import qualified Wuss
 
+-- | Activate the new deployment.
+--
+-- If the target profile is already locked by another deployment, exit
+-- immediately and rely on the backend to reschedule.
 main :: IO ()
 main = do
   setLocaleEncoding utf8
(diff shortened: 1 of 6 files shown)


cachix/cachix-deployment/Main.hs
```diff
@@ -30,20 +32,29 @@ import Protolude.Conv
...
 main = do
   setLocaleEncoding utf8
   hSetBuffering stdout LineBuffering
   hSetBuffering stderr LineBuffering
+
   input <- escalateAs (FatalError . toS) . Aeson.eitherDecode . toS =<< getContents
-  CachixWebsocket.runForever (CachixWebsocket.websocketOptions input) (handleMessage input)
+  let profile = toS . CachixWebsocket.profile . Input.websocketOptions $ input
+  void $
+    Lock.withTryLock ("/nix/var/nix/profiles/" <> profile) $
```
sandydoo, line 48: We should have the agent figure out the type of deployment (nixos, nix-darwin, etc) and provide this info to the deployer.

cachix/src/Cachix/Deploy/Lock.hs
```diff
@@ -0,0 +1,56 @@
...
+
+-- | Get the path to the lock directory based on the ownership of the profile.
+--
+-- For user-owned profiles: $XDG_CACHE_DIR/cachix/deploy/locks
+-- For root-owned profiles: /var/run/cachix/deploy/locks
+getLockDirectoryFromProfile :: FilePath -> IO FilePath
+getLockDirectoryFromProfile profilePath = do
+  userId <- Files.fileOwner <$> Files.getFileStatus (FilePath.takeDirectory profilePath)
+  if isRoot userId
+    then pure $ "/var/run" </> defaultLockDirectory
```
domenkozar, line 22: In order to minimize burden of support, it's best to always use the same path regardless of the user.

If we don't specialize `/var/run` and just use xdg's `/root/.cache` case that should work well, what do you think?

cachix/src/Cachix/Deploy/OptionsParser.hs
```diff
@@ -27,7 +27,7 @@ parser =
 
 data AgentOptions = AgentOptions
   { name :: Text,
-    profile :: Text
+    profile :: Maybe Text
```
domenkozar, line 30: :+1:

cachix/src/Cachix/Deploy/Lock.hs
```diff
@@ -0,0 +1,56 @@
...
+    Directory.emptyPermissions
+      & Directory.setOwnerReadable True
+      & Directory.setOwnerWritable True
+      & Directory.setOwnerExecutable True
+      & Directory.setOwnerSearchable True
+
+  let profile = FilePath.takeFileName profilePath
+  let lockFile = lockDirectory </> profile <.> "lock"
+
+  bracket
```
domenkozar, line 49: It's a bit disappointing that lukko doesn't offer bracket style API, but here we are :)

domenkozar (changes requested):

cachix/src/Cachix/Deploy/Lock.hs
sandydoo: I had no idea root had it’s own little space there 😂 That‘s awesome!

What happens on macOS though? Running `getXdgDirectory` with sudo, it returns the user cache directory and I end up creating a folder owned by root.

domenkozar: That might be because you need `sudo -H`, here's a session from m1 macos:

```
Prelude> import qualified System.Directory as Directory
Prelude Directory> Directory.getXdgDirectory Directory.XdgCache ""
"/var/root/.cache"
Prelude Directory> 
```

sandydoo: Ah, this again. I assume this works when running via launchd, but do we do if the user runs the agent manually? Show a warning if `$HOME` is incorrect?

domenkozar: Good question, a warning seems like a good idea.

domenkozar:
The only thing I'm worried about here is if the deployment binary doesn't exit for any reason, it will be just stuck. Any ideas how to guard against that?

sandydoo:
> The only thing I'm worried about here is if the deployment binary doesn't exit for any reason, it will be just stuck. Any ideas how to guard against that?

We try not to do that 🙃 

We could store the PID of the process together with some value that monotonically increases with each deployment. The agent could then check the lock file, look at the deployment data, and decide to kill the existing process if a new deployment comes in. Could be a JSON file. Could be SQLlite (if the agent will have more query-able data to store in the future, like logs).

domenkozar:
Let's leave that out of scope for now then, hopefully we can eliminate such bugs.

cachix/cachix-deployment/Main.hs
```diff
@@ -9,7 +9,9 @@ import Cachix.API.Error (escalateAs)
 import qualified Cachix.API.WebSocketSubprotocol as WSS
 import Cachix.Client.Retry
 import qualified Cachix.Deploy.Activate as Activate
+import qualified Cachix.Deploy.Lock as Lock
 import qualified Cachix.Deploy.Websocket as CachixWebsocket
+import qualified Cachix.Deploy.Websocket as Input
```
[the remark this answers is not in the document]
sandydoo: Typo 😬

PR #440, opened 2022-08-27
Deploy: communicate closure size

domenkozar:
I wasn't able to make `shellOut` to return `stdout` so I create a similar function with a slightly different behaviour :sweat_smile: 

While I was at it, I refactored all cache setup into a separate function.

### DIFF
--- a/cachix/src/Cachix/Deploy/Activate.hs
+++ b/cachix/src/Cachix/Deploy/Activate.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE DuplicateRecordFields #-}
 
 module Cachix.Deploy.Activate where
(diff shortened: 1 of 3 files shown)


cachix/src/Cachix/Deploy/Activate.hs
```diff
@@ -112,16 +88,21 @@ activate options connection sourceStream deploymentDetails agentInfo agentToken
           if not (all (== ExitSuccess) exitCodes)
             then deploymentFailed
             else do
-              now <- liftIO getCurrentTime
-              sendMessage $ deploymentFinished True now
+              now2 <- liftIO getCurrentTime
```
sandydoo, line 91: 🙈

cachix/src/Cachix/Deploy/Activate.hs
```diff
@@ -8,11 +9,18 @@ import qualified Cachix.Client.NetRc as NetRc
 import qualified Cachix.Deploy.Websocket as CachixWebsocket
 import qualified Cachix.Types.BinaryCache as BinaryCache
 import Cachix.Types.Permission (Permission (..))
+import qualified Data.Aeson as Aeson
+#if MIN_VERSION_aeson(2,0,0)
```
sandydoo, line 13: Do we have to support other versions of aeson?

domenkozar: The backend is still running aeson-1

sandydoo: Ah, I see! It consumes this as a library.

PR #443, opened 2022-09-04
deploy: rollback to previously known running profile

domenkozar:


sandydoo (changes requested):
What happens if the profile doesn’t exist? I could set the agent to manage a new profile. What are we rolling back to then?

domenkozar:
Good catch :roll_eyes: 

I've made the rollback optional based on the existance of the previous profile path.

sandydoo (approved):

PR #448, opened 2022-09-10
Refactor deployment

### LINKED ISSUE #441: Flush logs atomically
https://github.com/cachix/cachix/commit/cfa2949cfaff9769f39a1acad10d3422dbee4b25 addressed the queue bits, but we still need to make sure conduit finishes streaming logs and `sendMessage` command completes.

sandydoo:
#### Agent

- The websocket module is not longer in charge of setting up the agent and logging.
- :bug: Resolves an issue where the agent would queue and run several copies of the same deployment if the `StartDeployment` message was not sent quickly enough. The agent now spawns the deployment binary without blocking message processing. 

#### Deployment binary

- The deployment receives all of the information it needs for the deployment from the agent via stdin.
- :bug: The deployment no longer runs within the websocket message handler. Any errors within the socket won’t cause an infinite loop of socket reconnects.
- The logging and backend websockets run in their own separate threads.
- Activation now throws IO errors whenever any of the commands return a non-zero exit code.
- Logs and deployment commands are sent outside of activation, greatly simplifying data flow.

#### Websockets

- :bug: Queued messages are processed even when the websocket throws exceptions. See `receiveDataConcurrently`.
- :bug: Websockets are closed according to the spec. We send a `close` request and wait for a close request from the peer. The grace period for the response is 5 seconds, after which we close anyways. Resolves #441.
- ❗ Fixed a typo in the `receiveDataConcurrently` method.

#### Logging

- :star: The new `Log.withLog` method setups a Katip scribe and yields a ready-to-go logging context.
- Most functions are now plain IO. `KatipContextT` has been removed because it wasn’t being used. The result is: simpler typing and no `liftIO`.
- :star: `verbose` mode now mirrors the log stream to the local log.

## TODO

- [x] Rebase
- [x] Apply rollback and deployment fixes
- [x] Upgrade cachix-api on the backend to fix #441
- [x] Allow SSL-less connections for testing
- [ ] Bump cachix-api

### DIFF
--- a/cachix/src/Cachix/Deploy/StdinProcess.hs
+++ b/cachix/src/Cachix/Deploy/StdinProcess.hs
@@ -3,12 +3,12 @@ module Cachix.Deploy.StdinProcess where
 import Protolude hiding (stdin)
 import System.IO (hClose)
 import System.Process
-import Prelude (String, userError)
+import Prelude (String)
 
--- | Run a process with only stdin as an input
-readProcess :: FilePath -> [String] -> String -> IO ()
-readProcess cmd args input = do
-  (Just stdin, _, _, ph) <-
+-- | Spawn a process with only stdin as an input
+spawnProcess :: FilePath -> [String] -> String -> IO ()
+spawnProcess cmd args input = do
+  (Just stdin, _, _, _) <-
     createProcess
       (proc cmd args)
         { std_in = CreatePipe,
@@ -19,7 +19,3 @@ readProcess cmd args input = do
         }
   hPutStr stdin input
   hClose stdin
-  exitcode <- waitForProcess ph
-  case exitcode of
-    ExitSuccess -> return ()
-    ExitFailure code -> ioError $ userError $ "Process " <> show cmd <> " failed with exit code " <> show code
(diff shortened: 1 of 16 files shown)


cachix/src/Cachix/Deploy/StdinProcess.hs
```diff
@@ -19,7 +19,3 @@ readProcess cmd args input = do
         }
   hPutStr stdin input
   hClose stdin
-  exitcode <- waitForProcess ph
```
domenkozar, line 22: I don't quite understand how locking works if it's not waiting for the process to finish?

domenkozar:
@sandydoo anything left to do here?

cachix/src/Cachix/Deploy/StdinProcess.hs
sandydoo: This is, admittedly, a hack to consume the deployment messages eagerly, instead of accumulating them. While waiting for the binary to exit, the agent would continue to add deployments to the backlog. This can include duplicate deployments if the `startDeployment` message takes a while to be delivered.

Locking is done in the deployment binary (because the agent might be upgraded and lose the lock).

Waiting for the process is the right thing to do. With the current backend setup, I was thinking of tracking whether we're running a deployment and dropping any incoming deployments until the current one exits. But if the agent is restarted, we'll lose that state and we're back to this again.

PR #449, opened 2022-09-21
Allow configuring a persistent hostname

### LINKED ISSUE #439: config: specify hostname
Currently the cachix command allows specifying `--host` but it would be better if there was a command to make that permanent in cachix.dhall config file.

I was thinking like `cachix hostname <..>` but maybe it's better to go with `cachix config hostname <..>`.

sandydoo:
You can now configure a persistent hostname with `cachix config set hostname`. 

A few questions to resolve:

- [x] `host` vs `hostname`
`hostname` seems to be the standard. We could allow providing `--hostname` (in addition to the current `--host`)
- [x] `config set hostname` vs `config hostname`
Would we want to have other subcommands in the future? `get` or `list`?

Resolves #439.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,13 +9,14 @@
 $ cachix --help
 https://cachix.org command line interface
 
-Usage: cachix [--host URI] [-c|--config CONFIGPATH] [-v|--verbose] 
+Usage: cachix [--hostname URI] [-c|--config CONFIGPATH] [-v|--verbose]
               (COMMAND | (-V|--version))
+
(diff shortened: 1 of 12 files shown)


sandydoo:
@domenkozar, I went with `hostname`, but the CLI will still accept `host` to remain backwards-compatible.
I also decided to keep the `get`/`set` verbs in the end. This design seems pretty standard among CLIs that have config options. It’s obvious, it’s explicit, and you get to see help text when you run `config hostname`.

PR #459, opened 2022-10-18
Add a bootstrap mode to help initially deploy agents

### LINKED ISSUE #455: Add an option to have the agent exit after running a single deployment
This is useful for bootstrapping NixOS and NixDarwin systems, where the system-configured agent takes over after the initial deployment.

sandydoo:
Resolves #455.

I'll need to thoroughly look over this tomorrow, but I'd love some feedback!

The main additions are:

* The agent now acquires a lock on the profile that it's managing. This uses the same lock design as the deployment binary.
* The agent writes its PID to the same directory as the lock file.
* The new `--bootstrap` mode runs deployments until it detects another Cachix Deploy agent managing the same profile. In this mode,  the agent does not acquire a lock on the profile. We poll the PID file to find the newly spawned agent and double check that the process name is correct. Perhaps we can skip the process name check and just see if the lock file is locked?

#### Sample log output
```
[2022-10-18 02:42:06][agent.agent][Debug][agent-1][PID 875][ThreadId 8][cachix-1.0.1-5LusKggLrbP2kqeM106WV9:Cachix.Deploy.Websocket src/Cachix/Deploy/Websocket.hs:257:13] Received close request from peer (Peer initiated a close request). Closing connection
[2022-10-18 02:42:06][agent.agent][Info][agent-1][PID 868][ThreadId 4][cachix-1.0.1-5LusKggLrbP2kqeM106WV9:Cachix.Deploy.Agent src/Cachix/Deploy/Agent.hs:186:13] Waiting for another agent to take over the system profile...
[2022-10-18 02:42:06][agent.agent][Info][agent-1][PID 868][ThreadId 4][cachix-1.0.1-5LusKggLrbP2kqeM106WV9:Cachix.Deploy.Agent src/Cachix/Deploy/Agent.hs:197:17] Found an active agent for the system profile with PID 1045. Exiting.
```

#### Alternatives

I tried out a few other options before settling on the PID approach.

1. **`pgrep` with regex matching to find the profile in the arguments.** The main downside is that it ties the bootstrapping logic to the CLI design. I'm worried we might change the CLI and forget to update the regex. Also, it's regex 🤮.
2. **`ps`.** We have to extract and re-parse the command line arguments. This is better than `pgrep` because we can reuse our `OptionsParser` and avoid regex, but we still end up tied to the CLI.
3. **Manually reading proc files.** This isn't cross-platform, so it's best to stick with `ps`.

### DIFF
--- a/cachix/cachix-deployment/Main.hs
+++ b/cachix/cachix-deployment/Main.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE DuplicateRecordFields #-}
+{-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
 module Main
@@ -11,11 +12,11 @@ import qualified Cachix.API.WebSocketSubprotocol as WSS
 import qualified Cachix.Client.URI as URI
 import qualified Cachix.Deploy.Activate as Activate
 import qualified Cachix.Deploy.Agent as Agent
+import Cachix.Deploy.Deployment (Deployment (..))
 import qualified Cachix.Deploy.Lock as Lock
 import qualified Cachix.Deploy.Log as Log
 import qualified Cachix.Deploy.Websocket as WebSocket
 import qualified Control.Concurrent.Async as Async
-import qualified Control.Concurrent.MVar as MVar
 import qualified Control.Concurrent.STM.TMQueue as TMQueue
 import qualified Control.Exception.Safe as Safe
 import qualified Data.Aeson as Aeson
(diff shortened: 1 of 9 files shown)


domenkozar:
Nice :partying_face: I'd avoid wildcard extension, but if you feel like it helps I'm not strongly opinionated.

domenkozar:
> Perhaps we can skip the process name check and just see if the lock file is locked?

I'd go for skipping the process name and add it later if needed.

PR #465, opened 2022-10-21
Allow creating WebSocket state before connecting

sandydoo:
This is a minor improvement that lets you access the WebSocket record before making the connection. This turns out to be quite useful for setting up the WebSocket in a multi-threaded environment, where you'd want to set up the incoming message listeners before running the socket in a separate thread.

### DIFF
--- a/cachix/src/Cachix/Deploy/Websocket.hs
+++ b/cachix/src/Cachix/Deploy/Websocket.hs
@@ -45,6 +45,8 @@ import qualified Wuss
 data WebSocket tx rx = WebSocket
   { -- | The active WebSocket connection, if available
     connection :: MVar.MVar WS.Connection,
+    -- | The connection options
+    options :: Options,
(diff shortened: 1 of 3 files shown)


cachix/src/Cachix/Deploy/Websocket.hs
```diff
@@ -240,12 +248,8 @@ handleOutgoingJSON WebSocket {connection, tx} = do
 -- TODO: use exponential retry with reset: https://github.com/Soostone/retry/issues/25
 reconnectWithLog :: (MonadMask m, MonadIO m) => Log.WithLog -> m () -> m ()
 reconnectWithLog withLog inner =
-  Safe.handle closeRequest $
```
domenkozar, line 243: How come this isn't needed anymore?

sandydoo: Oh, shoot! Thanks, fixed!

PR #469, opened 2022-10-28
Add initial support for Home Manager

### LINKED ISSUE #422: Deploy: support home-manager
Raise your requests for home-manager support in this issue and I'll prioritize it :)

rycee:
This makes it possible to deploy standalone Home Manager configurations. Should work for any GNU/Linux distro on which Home Manager works.

Note, not fit for merge yet:

- [x] Test on a few GNU/Linux distros.
    - [x] Debian
    - [x] Ubuntu
    - [x] Fedora
- [x] Try out rollbacks.
- [x] Verify that NixOS deploys still work. Note, I don't have access to any macOS system so cannot verify nix-darwin deploys, but I imagine if NixOS still works then nix-darwin should as well?
- [x] Record brief demo screencast. See [Cachix demo on Ubuntu](https://walnut.rycee.net/cachix-demo.webm); sorry about the length and oddities in a few places.

Fixes #422

### DIFF
--- a/cachix/src/Cachix/Client/PushQueue.hs
+++ b/cachix/src/Cachix/Client/PushQueue.hs
@@ -127,7 +127,7 @@ exitOnceQueueIsEmpty stopProducerCallback pushWorker queryWorker queryWorkerStat
           cancelWith pushWorker StopWorker
         else do
           -- extend shutdown for another 90s
-          void $ Systemd.notify False $ "EXTEND_TIMEOUT_USEC=" <> show (90 * 1000 * 1000)
+          void $ Systemd.notify False $ "EXTEND_TIMEOUT_USEC=" <> show (90 * 1000 * 1000 :: Int)
           putTextError $ "Waiting to finish: " <> show inprogress <> " pushing, " <> show queueLength <> " in queue"
           threadDelay (1000 * 1000)
           go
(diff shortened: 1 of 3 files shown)


domenkozar:
Nice! Also needs an update of https://github.com/cachix/cachix/blob/master/cachix/src/Cachix/Deploy/OptionsParser.hs#L50

rycee:
Thanks 🙂

I've update `OptionsParser.hs` as well. I think the basic functionality is in place, I'll need to do a bit more testing and add some documentation to https://docs.cachix.org/deploy/.

rycee:
I'm pretty happy with the state right now. See the checklist in the PR description for things I've done. Just to double clarify, I've verified that NixOS deploys still work but I haven't done a check for macOS deploys due to not having such a system.

About the rollback script, it seems to work fine for HM activation but I guess if you have a deployment spec that includes both system and HM configurations you need to write the check script to suit both.

Log excerpt from a rollback run (on a Fedora system with Swedish locale):

```
⋮
Aktiverar onFilesChange
Aktiverar reloadSystemd
Stopping units: caffeine.service
Started org.gnome.Shell@x11.service - activating

Running rollback script.

$ /nix/store/2vwmk4lqvv1q64yz00jr71cmb3vvc26k-fail-deploy 
we'll fail this deploy

Deployment failed, rolling back ...

$ /nix/store/zmmrxw3g2vw8w8vc684f01j3fggcfqi1-home-manager-generation/activate 
Startar Home Manager-aktivering
Aktiverar checkFilesChanged
Aktiverar checkLinkTargets
⋮
```

PR #470, opened 2022-11-09
[#205] Add zstd compression to the binary cache

### LINKED ISSUE #205: Support brotli/zstd compression besides xz
In https://github.com/cachix/cachix/issues/55 it was pointed out that zstd/brotli both [can achieve much faster transfer speeds](https://community.centminmod.com/threads/round-3-compression-comparison-benchmarks-zstd-vs-brotli-vs-pigz-vs-bzip2-vs-xz-etc.17259/).

Player-205:
Problem: xz don't have sufficient transfer speed of binary cache

Solution: add zstd as another compression option to the push strategy

Fixes #205

### DIFF
--- a/cachix/src/Cachix/Client/Commands.hs
+++ b/cachix/src/Cachix/Client/Commands.hs
@@ -35,6 +35,7 @@ import Cachix.Client.Secrets
   )
 import Cachix.Client.Servant
 import qualified Cachix.Client.WatchStore as WatchStore
+import qualified Cachix.Types.BinaryCache as BinaryCache
 import qualified Cachix.Types.SigningKeyCreate as SigningKeyCreate
 import Control.Exception.Safe (throwM)
 import Control.Retry (RetryStatus (rsIterNumber))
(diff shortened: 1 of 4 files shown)


cachix/src/Cachix/Client/Commands.hs
```diff
@@ -224,11 +227,22 @@ getPushParams env pushOpts name = do
   pushSecret <- findPushSecret (config env) name
   store <- wait (storeAsync env)
   authToken <- Config.getAuthTokenMaybe (config env)
+  compressionMode <- case authToken of
```
domenkozar, line 230: This should rather case on `PushSecret` and disable this mode for when signing key is used.

Player-205: What the difference between auth token and push token in my case? AFAIU, I should just check PushSecret constructor and ignore its fields, right?

domenkozar: Almost zero, but you can have auth token and signing key, so it's best to look into `PushSecret` to make sure we cover both cases correctly. No need to look at the fields.

domenkozar (approved):
Could you squash please?

domenkozar:
Continuing in #472, thanks!

PR #472, opened 2022-12-14
Add zstd compression support

### LINKED ISSUE #205: Support brotli/zstd compression besides xz

domenkozar:
- [x] add compression mode to the cli

Fixes #205

### DIFF
--- a/cachix-api/src/Cachix/Types/BinaryCache.hs
+++ b/cachix-api/src/Cachix/Types/BinaryCache.hs
@@ -2,8 +2,9 @@ module Cachix.Types.BinaryCache where
 
 import Cachix.Types.Permission (Permission)
 import Data.Aeson (FromJSON, ToJSON)
-import Data.Swagger (ToSchema)
+import Data.Swagger (ToParamSchema (..), ToSchema)
 import Protolude
+import Servant.API
 
 data BinaryCache = BinaryCache
   { name :: Text,
@@ -11,6 +12,22 @@ data BinaryCache = BinaryCache
     isPublic :: Bool,
     publicSigningKeys :: [Text],
     githubUsername :: Text,
-    permission :: Permission
+    permission :: Permission,
+    compression :: CompressionMode
   }
   deriving (Show, Generic, FromJSON, ToJSON, ToSchema, NFData)
+
+data CompressionMode = XZ | ZSTD
+  deriving (Show, Read, Generic, FromJSON, ToJSON, ToSchema, NFData)
+
+instance FromHttpApiData CompressionMode where
+  parseUrlPiece "xz" = Right XZ
+  parseUrlPiece "zst" = Right ZSTD
+  parseUrlPiece compressionMode = Left $ "Wrong compression mode: " <> compressionMode
+
+instance ToHttpApiData CompressionMode where
+  toUrlPiece XZ = "xz"
+  toUrlPiece ZSTD = "zst"
+
+instance ToParamSchema CompressionMode where
+  toParamSchema _ = toParamSchema (Proxy :: Proxy Text)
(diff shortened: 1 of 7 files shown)


cachix-api/src/Cachix/Types/BinaryCache.hs
```diff
@@ -2,15 +2,32 @@ module Cachix.Types.BinaryCache where
...
 
 data BinaryCache = BinaryCache
   { name :: Text,
     uri :: Text,
     isPublic :: Bool,
     publicSigningKeys :: [Text],
     githubUsername :: Text,
-    permission :: Permission
+    permission :: Permission,
+    compression :: CompressionMode
```
sandydoo, line 16: Why does the binary cache report this? What is `compressionModeBackend` for?

sandydoo (changes requested):

cachix-api/src/Cachix/Types/BinaryCache.hs
domenkozar: There's a binary cache setting to pick what compression should the client use. That can be overriden with `--compression-mode` on the CLI.

sandydoo: Okay, so it's like a preferred encoding? But what's the intended use-case?

The way I see it, this remotely changes the client's default behavior, without actually enforcing that behavior. I worry that this may end up confusing users. Looking at this setting, I might ask:
* are all clients now forced to use this compression encoding?
* will all the NARs be stored in this encoding? Will they be re-encoded?
* will all the NARs be served in this encoding? 

Also,

1. How do we update the compression modes? Will we break older clients?
2. Should we instead let users permanently set the compression settings in their local config? They could either choose to use our defaults (ie, no preference), which we can change with new client releases, or they could request a specific mode (with levels, etc).

cachix/src/Cachix/Client/Commands.hs
```diff
@@ -214,7 +214,7 @@ pushStrategy store authToken opts name compressionMode storePath =
       onAttempt = \retrystatus size -> do
         path <- decodeUtf8With lenientDecode <$> storePathToPath store storePath
         -- we append newline instead of putStrLn due to https://github.com/haskell/text/issues/242
-        putStr $ retryText retrystatus <> "compressing and pushing " <> path <> " (" <> humanSize (fromIntegral size) <> ")\n",
+        putStr $ retryText retrystatus <> "compressing using " <> T.toLower (show compressionMode) <> " and pushing " <> path <> " (" <> humanSize (fromIntegral size) <> ")\n",
```
sandydoo, line 217: Could we show the compression settings once at the very top?

domenkozar: The level or did you have something else in mind?

sandydoo: Instead of printing out the compression mode for every path, we could print out the compression mode (and any other useful info) at the top. Then this line becomes much less verbose: `Pushing <path> (<size>)`.

sandydoo (approved):
👍

steveeJ:
this could be the origin of an issue that just was brought up to me by @zippy

```
copying path '/nix/store/8vqdszj1zgpv9q1j19pdqw9h76v5m2cc-rust-overlay-e4e129c' from 'https://holochain-ci.cachix.org'...
unknown compression method 'zstd'
error: build of '/nix/store/bs024sxrdzb96gasdk8llkhx3haj31dp-rust-overlay-e4e129c.drv' failed
(use '--show-trace' to show detailed location information)
```

@domenkozar please shed some light on this or tell me whether my suspicion could be accurate :pray:

domenkozar:
You need to upgrade Nix, you're using 2.3 or older.

PR #476, opened 2022-12-18
Flakeify

domenkozar:


stack.yaml
```diff
@@ -1,12 +1,12 @@
 # Match resolver with nixos-unstable haskellPackages
-resolver: lts-19.25
+resolver: lts-20.4
```
sandydoo, line 2: Should we stick with 19.33? GHC 9.0.2 is still the default on nixpkgs.

domenkozar: It was just merged today to be updated to lts-20

sandydoo: Awesome!

PR #482, opened 2022-12-28
Stream logs during agent deployment

sandydoo:


domenkozar:
We need to still support async method of deploying, for example when deploying a laptop, otherwise it will hang the activate deployment if the agent is offline.

I'm thinking naming it `activate --async` or without using too technical terms `activate --dont-wait`

domenkozar (approved):

PR #519, opened 2023-03-18
Hold main thread while remaining paths are pushed

sandydoo:
Related to https://github.com/NixOS/nixpkgs/pull/221167#issuecomment-1474835970.


cachix/src/Cachix/Client/Commands.hs
```diff
@@ -202,25 +202,26 @@ watchExec env pushOpts name cmd args = withPushParams env pushOpts name $ \pushP
         hDuplicateTo stderr stdout -- redirect all stdout to stderr
         WatchStore.startWorkers (pushParamsStore pushParams) (numJobs pushOpts) pushParams
 
-  Async.withAsync watch $ \watchThread -> do
-    -- Throw any errors encountered by the workers
-    Async.link watchThread
+  (_, exitCode) <-
+    Async.concurrently watch $ do
```
sandydoo, line 206: A lot of noise here, but the main change is the `concurrently`. The `wait` inside is crucial when using signals to shut things down.

Issue #374, opened 2021-05-13
Provide a statically linked binary


simonvpe:
I'm trying to optimize my build containers that I run on gitlab.com but just adding cachix to the image increases the image size by ~4GB. I've tried manually walking the dependencies for cachix and bringing in exactly the right shared obejcts it needs but when I do that I'm getting the error below.

I've made sure I can `curl` to 'https://cachix.org` and now I'm pretty much stuck. It'd be pretty sweet to have a statically linked binary that doesn't bring in the whole universe just for being able to upload a few binaries to a server.

```
bash-4.4# cachix use leash-lang-leash
warning: the group 'nixbld' specified in 'build-users-group' does not exist
cachix: ConnectionError (HttpExceptionRequest Request {
  host                 = "cachix.org"
  port                 = 443
  secure               = True
  requestHeaders       = [("Accept","application/json;charset=utf-8,application/json"),("Authorization","<REDACTED>"),("User-Agent","cachix 0.6.0")]
  path                 = "/api/v1/cache/leash-lang-leash"
  queryString          = ""
  method               = "GET"
  proxy                = Nothing
  rawBody              = False
  redirectCount        = 10
  responseTimeout      = ResponseTimeoutDefault
  requestVersion       = HTTP/1.1
}
 (ConnectionFailure Network.Socket.getAddrInfo (called with preferred socket type/protocol: AddrInfo {addrFlags = [AI_ADDRCONFIG], addrFamily = AF_UNSPEC, addrSocketType = Stream, addrProtocol = 0, addrAddress = <assumed to be undefined>, addrCanonName = <assumed to be undefined>}, host name: Just "cachix.org", service name: Just "443"): does not exist (No address associated with hostname)))
```

I'm using the flake btw, when installing from `nix-env -iA cachix -f https://cachix.org/api/v1/install` it seems to work better but that's not something that fits me very well since I want my containers to be pre-built when I launch them.

domenkozar:
Duplicates #184 

```
$ du -shc $(nix-store -qR /nix/store/q0lqpbbxcn2y4f23nhms7ncj7gvwbid1-cachix-0.6.0/bin/cachix)
...
150M	total
```

It's 150M, but that's mostly Nix also in the closure so if you install Nix from the same commit as Cachix it should really be about 50M extra.

[...]

Issue #475, opened 2022-12-17
Cachix leaves Nix DB broken inside Nix-Portable


Kha:
This is with Cachix 1.0.1 from Nixpkgs.

I'm experimenting with https://github.com/DavHau/nix-portable and trying to use `cachix push` from inside it. Even though the executable immediately exits because I forgot the auth token, it somehow modifies `/nix/var/nix/db/db.sqlite`, breaking any subsequent Nix invocations.
```
$ wget https://github.com/DavHau/nix-portable/releases/download/v009/nix-portable
$ chmod u+x nix-portable
$ ./nix-portable nix shell nixpkgs#cachix -c sh -c 'ls -l --full-time /nix/var/nix/db/db.sqlite; cachix push -j4 lean4 github:leanprover/lean4#Init; ls -l --full-time /nix/var/nix/db/db.sqlite'
Installing git. Disable this by specifying the git executable path with 'NP_GIT'
warning: cannot read flake registry '/home/sebastian/.config/nix/registry.json': error: opening file '/home/sebastian/.config/nix/registry.json': No such file or directory
-rw-r--r--    1 sebastia users      2498560 2022-12-17 16:46:38 +0000 /nix/var/nix/db/db.sqlite

Neither auth token nor signing key are present.

They are looked up via $CACHIX_AUTH_TOKEN and $CACHIX_SIGNING_KEY,
and if missing also looked up from ~/.config/cachix/cachix.dhall

Read https://mycache.cachix.org for instructions how to push to your binary cache.-rw-r--r--    1 sebastia users      2498560 2022-12-17 16:46:39 +0000 /nix/var/nix/db/db.sqlite
 $ ./nix-portable nix shell nixpkgs#cachix -c sh -c 'ls -l --full-time /nix/var/nix/db/db.sqlite; cachix push -j4 lean4 github:leanprover/lean4#Init; ls -l --full-time /nix/var/nix/db/db.sqlite'
error: executing SQLite statement 'pragma synchronous = normal': unable to open database file (in '/nix/var/nix/db/db.sqlite')
```
While this is a very fringe use case, it seems like this may point to a serious bug. From an strace log, it looks like the process exits while a thread is still modifying the DB?
```
...
[pid 2428400] pwrite64(41, "SQLite format 3\0\20\0\2\2\0@  \0\0\10\35\0\0\1\333"..., 4096, 0) = 4096
[pid 2428400] fdatasync(41)             = 0
[pid 2428400] close(42)                 = 0
[...]

PR #521, opened 2023-03-22
Rewrite C++ bits to Haskell

### LINKED ISSUE #184: Provide a static build?
Each of my travis builds now starts with
```
$ nix-env -iA cachix -f https://cachix.org/api/v1/install
unpacking 'https://cachix.org/api/v1/install'...
installing 'cachix-0.2.0'
these paths will be fetched (9.79 MiB download, 46.36 MiB unpacked):
  /nix/store/681354n3k44r8z90m35hm8945vsp95h1-glibc-2.27
  /nix/store/7mrq5v4nnqfkv1ka7a0kdw9mzvqs5w3c-gmp-6.1.2
  /nix/store/bnbq85lsbds5asi8vzjqhvrgq3l9ila8-cachix-0.2.0
  /nix/store/hlnxw4k6931bachvg5sv0cyaissimswb-gcc-7.4.0-lib
  /nix/store/iiymx8j7nlar3gc23lfkcscvr61fng8s-zlib-1.2.11
  /nix/store/k0y9nrym785l61pjw0lp0qk90v2nbpqs-ncurses-6.1-20190112
  /nix/store/kbbiky87bsdndh63m7919sxkf3gsh5zq-libffi-3.2.1
  /nix/store/mibh7947iwgf5yiqc7njn0igfzlxzdf1-xz-5.2.4
  /nix/store/y1ck4skkvmcnar0y0i1kkmvv0lasvy5b-cachix-0.2.0-data
copying path '/nix/store/y1ck4skkvmcnar0y0i1kkmvv0lasvy5b-cachix-0.2.0-data' from 'https://cache.nixos.org'...
copying path '/nix/store/681354n3k44r8z90m35hm8945vsp95h1-glibc-2.27' from 'https://cache.nixos.org'...
copying path '/nix/store/hlnxw4k6931bachvg5sv0cyaissimswb-gcc-7.4.0-lib' from 'https://cache.nixos.org'...
copying path '/nix/store/kbbiky87bsdndh63m7919sxkf3gsh5zq-libffi-3.2.1' from 'https://cache.nixos.org'...
copying path '/nix/store/7mrq5v4nnqfkv1ka7a0kdw9mzvqs5w3c-gmp-6.1.2' from 'https://cache.nixos.org'...
copying path '/nix/store/k0y9nrym785l61pjw0lp0qk90v2nbpqs-ncurses-6.1-20190112' from 'https://cache.nixos.org'...
copying path '/nix/store/mibh7947iwgf5yiqc7njn0igfzlxzdf1-xz-5.2.4' from 'https://cache.nixos.org'...
copying path '/nix/store/iiymx8j7nlar3gc23lfkcscvr61fng8s-zlib-1.2.11' from 'https://cache.nixos.org'...
copying path '/nix/store/bnbq85lsbds5asi8vzjqhvrgq3l9ila8-cachix-0.2.0' from 'https://cache.nixos.org'...
building '/nix/store/rrdzli02y27g1y4ld8ckgsk3f66yrvay-user-environment.drv'...
created 19 symlinks in user environment
The command "nix-env -iA cachix -f https://cachix.org/api/v1/install" exited with 0.
```
ok, it is only 15 seconds, but still: Have you considered building `cachix` statically, maybe with `musl`, i.e. `crossSystem = localLib.systems.examples.musl64`, so that this step has to download only one self-contained and hopefully smaller store path?

domenkozar:
Fixes #184 #374


roberth:
Oh wow I was going to send some warning fixes because I'd started to include cachix in my local build for tooling reasons. Seems that I may not have to bother with that? Anyway, that doesn't really matter, but...

Heaving read through the diff I'm concerned that doing it this way could cause trouble for you and your users. For instance, if you don't get the sqlite flags from the nix config correctly, you'll trigger #475 again. Also, you'll have to deal with changes to the db schema yourself. Probably other unknowns later.
Meanwhile, we're making progress on language bindings in nix upstream. A contributor is committing to C and Python bindings, so we'll have a good foundation for hsc2hs based bindings: no more template haskell and no more C++ interactions either.
Pick your poison :)

domenkozar:
Seems like we need to pick the right vfs: https://github.com/NixOS/nix/blob/master/src/libstore/sqlite.cc#L58

I'm not happy about this direction, but we had way too much trouble with C++ and this part of the codebase seems quite stable for now. In the future we can reverse this direction if the bindings become more sane.

cachix/src/Cachix/Client/Store.hs
```diff
@@ -0,0 +1,127 @@
...
+    go [SQLite.SQLText path_] = return path_
+    go a = throwIO $ FatalError $ "invalid reference type " <> show a
+
+    getDeriver :: SQLite.SQLData -> Maybe Text
+    getDeriver (SQLite.SQLText deriver) = Just deriver
+    getDeriver _ = Nothing
+
+computeClosure :: Store -> [StorePath] -> IO [StorePath]
+computeClosure store initialPaths = do
+  allPaths <- processGraph (getPath <$> initialPaths) (fmap (concatMap references . rights . List.singleton) . queryPathInfo store)
```
sandydoo, line 102: Do we care about any of the errors returned by `queryPathInfo` here? If not, should we document the reason?

What if we avoid going through a list here? The `.` could also be a `<$>` to be read as the common double fmap pattern.
```suggestion
  let fetchReferences = fmap (either (const []) references) . queryPathInfo store
  allPaths <- processGraph (getPath <$> initialPaths) fetchReferences
```

Or maybe even something really obvious:
```suggestion
  allPaths <- 
    processGraph (getPath <$> initialPaths) $
      queryPathInfo store >=> \case
        Left _ -> pure []
        Right pathInfo -> pure $ references pathInfo
```

cachix/src/Cachix/Client/Store.hs
```diff
@@ -0,0 +1,127 @@
...
+import qualified Data.Text as T
+import Database.SQLite3 (SQLData)
+import qualified Database.SQLite3 as SQLite
+import Protolude hiding (toS)
+import Protolude.Conv
+import System.Directory (canonicalizePath)
+import qualified System.Nix.Base32
+import System.Process (readProcessWithExitCode)
+
+data Store = Store Text SQLite.Database
```
sandydoo, line 18: We should document that the `Text` is a store prefix. Maybe a type alias?

sandydoo (changes requested):
👌

cachix/src/Cachix/Client/Store.hs
```diff
@@ -99,23 +102,23 @@ queryPathInfo store path = do
...
 
 getPath :: StorePath -> Text
 getPath (StorePath storePath) = storePath
 
-getStorePathBaseName :: StorePath -> Text
-getStorePathBaseName (StorePath storePath) =
-  T.drop 11 storePath
+getStorePathBaseName :: Store -> StorePath -> Text
+getStorePathBaseName (Store storePrefix _) (StorePath storePath) =
+  T.dropPrefix storePrefix storePath
```
sandydoo, line 123: The default prefix is `/nix`, so this isn't dropping the `/store/` bit. We might also have to sanitize the prefix to handle any trailing slashes.

PR #525, opened 2023-04-04
Open new store connection for each push

domenkozar:
This commit introduces a change to open a new connection to the store for each push operation. Previously, a single connection was opened at the beginning of a Cachix invocation, leading to issues when reading the SQLite database in immutable mode.

For instance, long-running commands like watch-* would not recognize new valid paths, as the database would not contain the new entries.

To address this issue, we now open a new database connection whenever the push logic is invoked.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,3 +22,5 @@ jobs:
     - run:  nix build -L '.#ci' '.#cachix' --show-trace
     # make sure it's all uploaded to cachix
     - run: echo > /tmp/store-path-pre-build
+    # test watch-exec
+    - run: cachix watch-exec cachix -- nix-build genpaths.nix --substituters 'https://cache.nixos.org'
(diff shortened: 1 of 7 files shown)


sandydoo (commented):
I still don't fully understand the implications of 'immutable', but I guess there's no other choice if we're sticking with this approach.

I'm also a bit concerned with the overhead this may introduce per store path. ~Our CI run took 40% longer to push the relatively few paths it normally does.~ What will happen in larger CI runs?

My only other comment would be to point out that this "direct to db" approach is a dead-end  for us. I'll admit, it was a great idea! In practice, the db schema is far more stable than the libstore for our narrow use-case. But the unfortunate realities of trying to read the db, particularly in a multi-user setup, make this a no-go for me in the long run. I hope a C ABI to the store can be agreed upon soon!

domenkozar:
Benchmarking 500 paths of 1MB shows no regressions (actually a percent or two of improvement) and running a bunch of builders in parallel with the same store path shows it works well.

PR #530, opened 2023-04-22
Store: don't overresolve symlinks that are also store paths

K900:
It's possible for /nix/store/aa...-foo to be a symlink to /nix/store/aa...-bar. In this case, don't follow the symlink when normalizing the path, otherwise we end up getting the data for the path being linked to.


K900:
Closed as per discussion with Domen, who has a different solution incoming.

PR #534, opened 2023-05-12
Pins

domenkozar:


roberth:
A call to `fetchFromGitHub` is also called a _pin_.
I suppose _root_ would be good for the GC part, but not cover the feature. I see an undocumented `artifacts` attribute in there.
How about calling these _release_ instead?

domenkozar:
> A call to `fetchFromGitHub` is also called a _pin_. I suppose _root_ would be good for the GC part, but not cover the feature. I see an undocumented `artifacts` attribute in there. How about calling these _release_ instead?

There are many mentions of root in terms of Nix too, we have picked a name that we think is better suited for users.

We have a PR open with the documentation, should be merged soon.

PR #542, opened 2023-06-12
Add a default mkShell dev shell

sandydoo:
Stack has trouble finding external libraries in devenv. This is an escape hatch for now.


sandydoo:
Oh, the `watch-exec` test will now fail for all PRs from forks.

sandydoo:
I've disabled the watch-exec test for runs without the signing key.

PR #546, opened 2023-06-20
Display a progress bar for each path while pushing

domenkozar:


sandydoo (changes requested):

sandydoo:
@domenkozar, the one question I have is what this looks like in CI? Should progress bars be disabled for non-TTY outputs?

domenkozar:
@sandydoo fixed :)

cachix/src/Cachix/Client/Push.hs
```diff
@@ -138,6 +140,18 @@ getCacheAuthToken :: PushSecret -> Token
 getCacheAuthToken (PushToken token) = token
 getCacheAuthToken (PushSigningKey token _) = token
 
+updateProgressBar :: MonadIO m => Maybe ProgressBar -> ConduitM ByteString ByteString m ()
+updateProgressBar Nothing = awaitForever Data.Conduit.yield
+updateProgressBar (Just pg) =
+  await
+    >>= maybe
```
roberth, line 147: iiuc this is a `let m = await >>= maybe pass m`, and that means it's another `awaitForever`.

cachix/src/Cachix/Client/Push.hs
```diff
@@ -172,14 +186,15 @@ uploadStorePath cache storePath retrystatus = do
...
-  onAttempt strategy retrystatus storePathSize
+  maybeProgressBar <- onAttempt strategy retrystatus storePathSize
 
   withCompressor $ \compressor -> liftIO $ do
     uploadResult <-
       runConduitRes $
         streamNarIO narEffectsIO (toS storePathText) Data.Conduit.yield
           .| passthroughSizeSink narSizeRef
           .| passthroughHashSink narHashRef
+          .| updateProgressBar maybeProgressBar
```
roberth, line 197: Make this more general to achieve better separation of concerns. See comment in `PushStrategy`.

```suggestion
          .| spyIntoNarStream retrystatus storePathSize
```

cachix/src/Cachix/Client/Push.hs
```diff
@@ -85,7 +87,7 @@ data PushParams m r = PushParams
 data PushStrategy m r = PushStrategy
   { -- | Called when a path is already in the cache.
     onAlreadyPresent :: m r,
-    onAttempt :: RetryStatus -> Int64 -> m (),
+    onAttempt :: RetryStatus -> Int64 -> m (Maybe ProgressBar),
```
roberth, line 90: This gives a better separation of concerns, not expanding the scope of this API client module into CLI concerns.

```suggestion
    onAttempt :: RetryStatus -> Int64 -> m (),
    -- | A conduit to interject between NAR generation and compression. This allows the caller to perform a side effect during streaming, e.g. to update a progress bar. This conduit must the NAR bytestream through unmodified.
    spyIntoNarStream :: RetryStatus -> Int64 -> ConduitM ByteString ByteString m (),
```

Besides better separation of concerns, this solution is a bit more useful because it allows the caller to perform a side effect both before and after the `yield` (which I think is meaningful wrt output blocking, but not 100% sure), and it could perhaps even be used for other purposes, such as producing a listing of the directory contents in the same pass. I don't know how relevant those use cases are, but you get them for free.

cachix/src/Cachix/Client/Push.hs
```diff
@@ -138,6 +140,18 @@ getCacheAuthToken :: PushSecret -> Token
...
 
+updateProgressBar :: MonadIO m => Maybe ProgressBar -> ConduitM ByteString ByteString m ()
+updateProgressBar Nothing = awaitForever Data.Conduit.yield
+updateProgressBar (Just pg) =
+  await
+    >>= maybe
+      (return ())
+      ( \chunk -> do
+          liftIO $ tickN pg $ BS.length chunk
+          Data.Conduit.yield chunk
```
roberth, line 151:
```suggestion
          Data.Conduit.yield chunk
          liftIO $ tickN pg $ BS.length chunk
```

It's only done when it's done. That's assuming the return of `yield` represents completion of that chunk. Probably still not entirely accurate because there's going to be some buffered but unsubmitted input until you flush the compressed stream, which I don't think is something you'd want to do. Anyway, the suggestion seems slightly better.

roberth (changes requested):
Some suggestions to improve responsibilities and keep `Client.Push` flexible and side effect free for library users who want to implement their own progress updates.

If you want library users to be able to reuse the default progress updater, your command line progress bar, I would recommend to put all of its code into a separate module in the library. If that's not the case, it's probably still nice to have it in a separate module in the `cachix:exe` component of the package.

domenkozar:
@roberth wdyt?

sandydoo (approved):

cachix/src/Cachix/Client/Push.hs
```diff
@@ -83,31 +84,38 @@ data PushParams m r = PushParams
...
 
 data PushStrategy m r = PushStrategy
-  { -- | Called when a path is already in the cache.
+  { -- | An action to call when a path is already in the cache.
     onAlreadyPresent :: m r,
+    -- | An action to run on each push attempt.
     onAttempt :: RetryStatus -> Int64 -> m (),
+    -- | A conduit to inspect the NAR stream before compression.
+    -- This is useful for tracking progress and computing hashes. The conduit must not modify the stream.
+    onStream :: RetryStatus -> Int64 -> ConduitT ByteString ByteString (ResourceT m) (),
```
roberth, line 93:
```suggestion
    onUncompressedNARStream :: RetryStatus -> Int64 -> ConduitT ByteString ByteString (ResourceT m) (),
```

Preemptive disambiguation for when you want to show the network and/or compression speed (whichever is the bottleneck), using the compressed stream.

cachix/src/Cachix/Client/Commands.hs
```diff
@@ -256,18 +260,41 @@ retryText :: RetryStatus -> Text
...
+              def
+                { pgTotal = fromIntegral size,
+                  -- https://github.com/yamadapc/haskell-ascii-progress/issues/24
+                  pgWidth = 20 + barLength,
+                  pgOnCompletion = Just $ color Green "✓ " <> toS path <> " (" <> hSize <> ")",
+                  pgFormat = bar
+                }
+
+        isTerminal <- liftIO $ hIsTerminalDevice stdout
+        unless isTerminal $ do
```
roberth, line 290: I would expect an `if isTerminal` here; two branches, instead of doing the progress bar even when there's no terminal.

cachix/src/Cachix/Client/Commands.hs
```diff
@@ -256,18 +260,41 @@ retryText :: RetryStatus -> Text
 retryText retrystatus =
   if rsIterNumber retrystatus == 0
     then ""
-    else "(retry #" <> show (rsIterNumber retrystatus) <> ") "
+    else color Yellow $ "retry #" <> show (rsIterNumber retrystatus) <> " "
```
roberth, line 263: Does this account for having no color support?

roberth (commented):
Happy about the new push library function; just the name nitpick.

Not sure about how you handle the case where users don't want color or when they have a dumb terminal, but I'm going to trust that you're going to get that right. Maybe I just haven't read enough of the cachix CLI code to assess that properly.

cachix/src/Cachix/Client/Commands.hs
sandydoo: Yeah, me too.

cachix/src/Cachix/Client/Push.hs
sandydoo: I was thinking `onStream`/`onCompressedStream`, but this works too.

PR #548, opened 2023-06-23
cachix remove

### LINKED ISSUE #267: cachix CLI command for removing/disabling cache
It would be nice to have a `cachix` command for easily disabling or removing a cachix cache.

For instance, I added the `static-haskell-nix` cache:

```console
$ sudo cachix use static-haskell-nix
```

Later, I wanted to remove it, so I looked for a command like `cachix remove` or `cachix disable`, but it didn't exist.

I guess the command would simply just remove the `/etc/nixos/cachix/static-haskell-nix.nix` file (as well as reminding you to re-run `nixos-rebuild switch`?).

domenkozar:
Fixes #267 
```
❯ cachix use cachix
Configured https://cachix.cachix.org binary cache in /home/domen/.config/nix/nix.conf

❯ cachix remove cachix
Removed https://cachix.cachix.org binary cache in /home/domen/.config/nix/nix.conf
```

### DIFF
--- a/cachix/src/Cachix/Client/InstallationMode.hs
+++ b/cachix/src/Cachix/Client/InstallationMode.hs
@@ -5,11 +5,13 @@ module Cachix.Client.InstallationMode
     NixEnv (..),
     getInstallationMode,
     addBinaryCache,
+    removeBinaryCache,
     isTrustedUser,
     getUser,
     fromString,
     toString,
     UseOptions (..),
+    defaultUseOptions,
   )
 where
 
@@ -18,11 +20,13 @@ import qualified Cachix.Client.Config as Config
 import Cachix.Client.Exception (CachixException (..))
 import qualified Cachix.Client.NetRc as NetRc
 import qualified Cachix.Client.NixConf as NixConf
+import qualified Cachix.Client.URI as URI
 import qualified Cachix.Types.BinaryCache as BinaryCache
 import qualified Data.Maybe
 import Data.String.Here
 import qualified Data.Text as T
-import Protolude
+import Protolude hiding (toS)
+import Protolude.Conv (toS)
 import System.Directory (Permissions, createDirectoryIfMissing, getPermissions, writable)
 import System.Environment (lookupEnv)
 import System.FilePath (replaceFileName, (</>))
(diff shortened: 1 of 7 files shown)


cachix/src/Cachix/Client/InstallationMode.hs
```diff
@@ -122,18 +147,19 @@ addBinaryCache config bc _ (Install ncl) = do
...
+
+removeBinaryCache :: Text -> InstallationMode -> IO ()
+removeBinaryCache name (Install ncl) = do
+  (input, output) <- prepareNixConf ncl
+  let (final, removed) = NixConf.remove name input output
+  NixConf.write ncl final
   filename <- NixConf.getFilename ncl
-  putStrLn $ "Configured " <> BinaryCache.uri bc <> " binary cache in " <> toS filename
+  if removed
+    then putStrLn $ "Removed https://" <> name <> ".cachix.org binary cache in " <> toS filename
```
domenkozar, line 159: we should probably get the full uri from the config that was removed and show it here

cachix/src/Cachix/Client/NixConf.hs
```diff
@@ -94,6 +95,18 @@ add bc toRead toWrite =
...
 
+remove :: Text -> [NixConf] -> NixConf -> (NixConf, Bool)
+remove name toRead toWrite =
+  (newconf, oldsubstituters /= substituters)
+  where
+    newconf =
+      writeLines isPublicKey (TrustedPublicKeys $ nub publicKeys) $
+        writeLines isSubstituter (Substituters $ nub substituters) toWrite
+    oldsubstituters = readLines toRead isSubstituter
+    substituters = filter (not . T.isInfixOf ("//" <> name <> ".")) oldsubstituters
```
domenkozar, line 106: we should append the host config here

roberth:
> unuse

Unusual.

sandydoo (approved):
I added somewhat of a test 🙂 

Also, `removeBinaryCache` will always insist that it removed the cache whenever it's in the global conf and not the local one. The way `prepareNixConf` combines the configs is a bit painful!

PR #561, opened 2023-07-15
Filter invalid paths when watching the store

### LINKED ISSUE #547: `cachix watch-exec` fails with `CppStdException "path '/nix/store/... is not valid" (Just "nix::InvalidPath")`
I've only witnessed this bug when `/nix/store/60vgs5ip6bp7az83z7yi3hkrpgjcp2qd-libidn2-2.3.4` appeared in the Nix store while building.
Error output:
```
CppStdException "path '/nix/store/60vgs5ip6bp7az83z7yi3hkrpgjcp2qd-libidn2-2.3.4' is not valid" (Just "nix::InvalidPath")
```
I guess this is partially caused by this [no-bootstrap-reference hack](https://github.com/NixOS/nixpkgs/blob/26937f95223cdd917418dfc4e9957a5f81f012f9/pkgs/stdenv/linux/default.nix#L663).

Originally, this happened while building `systemd 253` with Cachix on newer nixpkgs.

### Reproduce
https://github.com/erikarvstedt/cachix-bug-repro builds a drv with `cachix watch-exec`, triggering the bug.
The error is visible in the [CI build log](https://github.com/erikarvstedt/cachix-bug-repro/runs/14534724721).

### Affected Cachix versions
Probably all recent versions (I've checked 1.1, 1.2, 1.3.3, 1.5).
In versions <=  1.2, Cachix exits, but the Nix build command keeps running, with output attached to TTY.

sandydoo:
It appears that Nix can produce "ghost paths" when a derivation's output is symlinked to another a path, like in the case of the referenced libidn derivation.

Even if this is a bug in Nix, this store path now exists (or doesn't) and we need to filter it out. In the case of the watch-style commands, we had filtering on the consuming side, but not on the producing/querying side. This would lead to a crash when computing the closure.

Fixes #547.

### DIFF
--- a/cachix/src/Cachix/Client/WatchStore.hs
+++ b/cachix/src/Cachix/Client/WatchStore.hs
@@ -3,6 +3,7 @@ module Cachix.Client.WatchStore
   )
 where
 
+import Cachix.Client.CNix (filterInvalidStorePath)
 import Cachix.Client.Push
 import qualified Cachix.Client.PushQueue as PushQueue
 import qualified Control.Concurrent.STM.TBQueue as TBQueue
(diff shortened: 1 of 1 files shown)


cachix/src/Cachix/Client/WatchStore.hs
```diff
@@ -25,7 +26,9 @@ producer store mgr queue = do
 queueStorePathAction :: Store -> PushQueue.Queue -> Event -> IO ()
 queueStorePathAction store queue (Removed lockFile _ _) = do
   sp <- Store.parseStorePath store (encodeUtf8 $ toS $ dropLast 5 lockFile)
-  atomically $ TBQueue.writeTBQueue queue sp
+  filterInvalidStorePath store sp >>= \case
+    Nothing -> return ()
```
domenkozar, line 30: We should issue a warning here

sandydoo: It prints out the usual warning that it skipped over the path.

PR #570, opened 2023-08-18
Add a daemon mode to push store paths asynchronously

sandydoo:
A basic daemon with muti-cache support.

Supports the following commands:
* `cachix daemon run` - Launches the daemon.
* `cachix daemon stop` - Stops the daemon and waits for it to finish pushing store paths.
* `cachix daemon push` - Queues store paths.

### DIFF
--- a/cachix/src/Cachix/Client/Commands.hs
+++ b/cachix/src/Cachix/Client/Commands.hs
@@ -309,7 +309,7 @@ pushStrategy store authToken opts name compressionMethod storePath =
             return $ liftIO . tickN progressBar . BS.length
           else do
             -- we append newline instead of putStrLn due to https://github.com/haskell/text/issues/242
-            putStr $ retryText retryStatus <> "Pushing " <> path <> " (" <> toS hSize <> ")\n"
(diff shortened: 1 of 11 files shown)


cachix/src/Cachix/Client/Commands.hs
```diff
@@ -308,7 +309,7 @@ pushStrategy store authToken opts name compressionMethod storePath =
             return $ liftIO . tickN progressBar . BS.length
           else do
             -- we append newline instead of putStrLn due to https://github.com/haskell/text/issues/242
-            putStr $ retryText retryStatus <> "Pushing " <> path <> " (" <> toS hSize <> ")\n"
+            putErrText $ retryText retryStatus <> "Pushing " <> path <> " (" <> toS hSize <> ")\n"
```
domenkozar, line 312: :partying_face:

sandydoo: Actually, maybe this isn't correct 🤔 `putErrText` uses `putStrLn` behind the scenes.

domenkozar: Most likely needs a `putErrStr` version (maybe even in protolude).

PR #598, opened 2023-11-10
dev: improve the release process

### LINKED ISSUE #489: git tag should contain the cabal bump

### LINKED ISSUE #537: Bump cabal before tagging
Two issues happen during release:

- bumping cabal later means anyone using the tag gets the wrong version
- there's no CI build for the commits done by github actions

sandydoo:
Run `./release.sh` to:
1. See the current package versions
2. Enter a new version
3. Update the package versions
4. Commit and add a version tag
5. Remind you to double-check and then push

Our CI then takes care of building, running tests, packaging the tarballs, and publishing.

Resolves #489.
Resolves #537.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,28 +1,31 @@
 name: "Release"
+
 on:
-  push:
-    tags:
-      - v**
+  workflow_call:
+    secrets:
+      HACKAGE_AUTH_TOKEN:
+        required: true
+
 jobs:
   release:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
-    - uses: cachix/haskell-release-action@v1
-      with:
-        branch: master
-        work-dir: cachix-api
-        hackage-token: ${{ secrets.HACKAGE_AUTH_TOKEN }}
-    - uses: cachix/haskell-release-action@v1
-      with:
-        branch: master
-        work-dir: cachix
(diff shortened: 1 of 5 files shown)


domenkozar:
I was thinking we could have two workflows:

1) triggered manually that sets the new version and commits, tags
2) triggers on tag, pushes to hackage (as it is now)

sandydoo:
> I was thinking we could have two workflows:
> 
>     1. triggered manually that sets the new version and commits, tags
> 
>     2. triggers on tag, pushes to hackage (as it is now)

That's what this does though. The only difference is that (1) is a local script and not an action. We could wrap it into a dispatchable action, but is there a substantial benefit to that?

.github/workflows/release.yml
```diff
@@ -1,28 +1,31 @@
 name: "Release"
+
 on:
-  push:
-    tags:
-      - v**
+  workflow_call:
```
domenkozar, line 4: `This used to trigger on tag push and release cachix-api/cachix, what's the process now? Is there something missing in `RELEASE.md`?

sandydoo: The `workflow_call` lets it be triggered by another workflow. This happens after our CI runs successfully and only on version tags. https://github.com/cachix/cachix/pull/598/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R41-R45

It's not as elegant as having the version tag trigger right here, but at least our releases get run through CI before being published.

PR #613, opened 2023-12-25
daemon: add a push manager with active store path deduping

sandydoo:

PR #616, opened 2023-12-27
daemon: support nix conf overrides

sandydoo:
Unlike `NIX_USER_CONF_FILES`, which concatenates multiple nix.confs, `NIX_CONF` completely overrides the config. This PR supports registering the daemon post-build-hook even if `NIX_CONF` is used.

Depends on #613.


domenkozar:
Needs a rebase

PR #641, opened 2024-05-09
daemon: lifecycle hardening

sandydoo:
- Refactor the main blocking thread into an event loop to make it possible to manage internal daemon state. For example, react to new messages, add and remove active sockets, etc.
- Add ping-pong messages to the protocol to allow for liveness checks.
- Remove socket conversion to blocking mode when connecting as a client. This causes hangs on some systems.


sandydoo:
Closing in favour of #647.

Issue #680, opened 2025-01-10
Support for nix.conf include directives


vilhelmbergsoe:
**Describe the bug**
When using determinate nix installer (and similar setups that use include directives), Cachix fails to recognize trusted-users settings from included config files.

Steps to reproduce:
1. Using determinate nix installer setup with:
   - /etc/nix/nix.conf containing `!include nix.custom.conf`
   - /etc/nix/nix.custom.conf containing `trusted-users = <user>`
2. Run `cachix use <cache>`

Current output:
```
This user doesn't have permissions to configure binary caches.

You can either:

a) Run the same command as root to configure them globally.

b) Run the following command to add your user as trusted
   and then try again:

  echo "trusted-users = root <user>" | sudo tee -a /etc/nix/nix.conf && sudo pkill nix-daemon
```

Expected behavior:
- Cachix should recognize trusted-users settings from included config files
- Both required (`include`) and optional (`!include`) directives should work
- Similar to how `nix config show` handles includes (which correctly shows the trusted-users setting)

**Version**
```
cachix 1.7.5
```

dzmitry-lahoda:
Just got it into our CI out of nowhere - no changes from our side.

dzmitry-lahoda:
Seems our case because of https://github.com/DeterminateSystems/nix-installer/releases/tag/v0.33.0

grahamc:
[moved from #681]

Hey folks, sorry about this breakage ... ack. It definitely wasn't on our radar that cachix was parsing the nix.conf and would get tripped up by this. 

We've rolled back the installer to the previous version for the moment to mitigate the issue.

Would it be sensible to update the nix.custom.conf if it exists and is included? For users of Determinate Nix (which this issue does not strictly involve), the nix.conf gets rewritten on startup, which will continue to break cachix users until the config is placed in the nix.custom.conf.

grahamc:
By way of update on our end: we're going to temporarily write the `trusted-users` setting, if specified, to /etc/nix/nix.conf as well as /etc/nix/nix.custom.conf when users install upstream Nix. This will unblock our release process and allow us to continue moving forward.

We made this split so nix.conf is predictable, and nix.custom.conf is inspectable for what the user set. We're going to continue working on this as part of our roadmap. Obviously we didn't think it would impact Cachix users.

Importantly, though:

1. Determinate Nix will continue having trusted-users in the nix.custom.conf for the foreseeable future,
[...]

PR #681, opened 2025-01-10
feat: add support for nix.conf include directives

vilhelmbergsoe:
Implement parsing and resolution of include/!include directives in nix.conf files to support setups like determinate nix installer that split configs across multiple files.

## Key features:

- Support both required (include) and optional (!include) directives
- Detect and prevent circular includes

## Implementation details:

Following the [nix.conf specification](https://nix.dev/manual/nix/2.17/command-ref/conf-file#file-format):
- Required includes (`include`) error if file not found
- Optional includes (`!include`) continue silently if file missing
- Circular includes are detected, with a stack trace showing the include chain, example:

```sh
Circular include detected:
/path/to/nix.conf
    -> includes /path/to/nix.custom.conf
    -> includes /path/to/nix.conf (circular reference)
```

- Optional includes that would create a circular reference are skipped silently
- All paths are resolved relative to the including file

This allows Cachix to properly read trusted-users and other settings from included config files.

---

I'm new to Haskell, so I'd really appreciate feedback on the implementation approach. Happy to adjust based on your input!

Related issue: #680

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -1,2 +1,3 @@
 - ignore: { name: Use newtype instead of data }
 - ignore: { name: Redundant do }
+- ignore: { name: "Eta reduce" }
--- a/cachix/cachix.cabal
+++ b/cachix/cachix.cabal
@@ -263,6 +263,7 @@ test-suite cachix-test
     , dhall
     , directory
     , extra
+    , filepath
     , hercules-ci-cnix-store
     , here
     , hspec
--- a/cachix/src/Cachix/Client/Exception.hs
+++ b/cachix/src/Cachix/Client/Exception.hs
@@ -12,6 +12,8 @@ data CachixException
   | NoInput Text
   | NoConfig Text
   | NetRcParseError Text
+  | IncludeNotFound Text
+  | CircularInclude Text
   | NarStreamingError ExitCode Text
   | NarHashMismatch Text
   | DeprecatedCommand Text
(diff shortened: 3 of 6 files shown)


sandydoo:
The parsing bits look good to me. I added tests for those and ran the formatters over the code. There are still some minor issues with variable shadowing.

The main problem is the write-back function. If we flatten all of the includes into a single `NixConf`, then we'll mangle the config when writing it back. 

I haven't given this much thought yet. Perhaps we should use our own `include` for stuff that needs to be written to the config. That

vilhelmbergsoe:
Thanks giving this a view!

If I understand correctly, the write function is only used for writing our own configuration changes, not modifying the user's existing config files. In that case, flattening should be fine since we're not trying to preserve/modify the original include structure.

Would you confirm that's the intended behavior of the write function? If so, I think the current flattening approach should work well for our needs.

grahamc:
Hey folks, sorry about this breakage ... ack. It definitely wasn't on our radar that cachix was parsing the nix.conf and would get tripped up by this. 

We've rolled back the installer to the previous version for the moment to mitigate the issue.

Would it be sensible to update the nix.custom.conf if it exists and is included? For users of Determinate Nix, the nix.conf gets rewritten on startup, which will continue to break cachix users until the config is placed in the nix.custom.conf.

sandydoo:
> If I understand correctly, the write function is only used for writing our own configuration changes, not modifying the user's existing config files.

It very much modifies the existing config files in-place.

I've worked out an interim solution for the writes: https://github.com/cachix/cachix/compare/nix-conf-fix?expand=1

@domenkozar and I will review tomorrow morning (it's quite late here), likely apply the above changes on top, and start propagating a release.

PR #696, opened 2025-07-11
Allow daemon socket file path to be read from `CACHIX_DAEMON_SOCKET`

joshuaspence:


sandydoo (changes requested):
A few things to fix, otherwise happy to merge!

PR #698, opened 2025-07-18
Make `cachix daemon push` synchronous

joshuaspence:


joshuaspence:
@sandydoo Thanks it looks good!

PR #704, opened 2025-08-28
Cachix.Client.NixVersion: disambiguate show

roberth:
Required for GHC 9.10

> Add Data.Text.show and Data.Text.Lazy.show

https://github.com/haskell/text/blob/master/changelog.md#212

Choice is `Protolude.show` so that we don't require such a recent `text`


sandydoo:
Thanks!

PR #723, opened 2026-03-10
client: support non-default store locations

sandydoo:
Queries the configured store path from Nix and uses it throughout the client.

Also exposes a new `--store` option to allow overriding the store path temporarily.


sandydoo:
Need to expose some APIs in cnix-store before this.

PR #728, opened 2026-04-09
push: exit when parent process dies to prevent orphaned spinners

danv-anduril:
When `cachix push` reads store paths from stdin (e.g. `nix-build | cachix push`), the parent process may be killed while uploads are in-flight. Because the HTTP client uses `responseTimeoutNone`, requests hang indefinitely on dead (CLOSE-WAIT) connections. The orphaned cachix process then spins at 100% CPU on GHC's IO manager timerfd, consuming a full core forever.

This was observed on CI runners where ~40 orphaned `cachix push` processes accumulated over hours, each burning 100% CPU with CLOSE-WAIT sockets and PPID 1.

Fix: when reading from stdin, fork a watchdog thread that polls `getParentProcessID` every 5 seconds. If the parent PID changes (process re-parented to init), exit cleanly.

Tested by:
- Building a 500MB random store path
- Starting `cachix push` in a pipe from a setsid'd parent
- Blocking outbound traffic with iptables to force CLOSE-WAIT
- Killing the parent process

Master cachix (e8be573): orphaned at PPID 1, 100% CPU, CLOSE-WAIT sockets, alive indefinitely.
Patched cachix: detected parent death, printed "cachix: parent process died, exiting", exited in <5 seconds.

### DIFF
--- a/cachix/src/Cachix/Client/Command/Push.hs
+++ b/cachix/src/Cachix/Client/Command/Push.hs
@@ -24,6 +24,7 @@ import Cachix.Client.Secrets
 import Cachix.Client.Servant
 import Cachix.Types.BinaryCache (BinaryCacheName)
 import Cachix.Types.BinaryCache qualified as BinaryCache
+import Control.Concurrent (forkIO, myThreadId, threadDelay, throwTo)
 import Control.Exception.Safe (throwM)
 import Control.Retry (RetryStatus (rsIterNumber))
 import Data.ByteString qualified as BS
@@ -33,7 +34,7 @@ import Data.Text qualified as T
 import Hercules.CNix (StorePath)
 import Hercules.CNix.Store (Store, storePathToPath, withStore)
 import Network.HTTP.Types (status401, status404)
-import Protolude hiding (toS)
+import Protolude hiding (toS, throwTo)
 import Protolude.Conv
 import Servant.Auth ()
 import Servant.Auth.Client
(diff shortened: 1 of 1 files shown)


sandydoo (commented):
I see two separate issues here, so lets talk through them separately.

### 1. What happens when the process is orphaned?

The parent process (shell, CI runner, etc) was killed for some reason (forcefully?), which left cachix running and re-parented it to PID 1.

Isn't this normal behavior? Are we expected to exit when re-parented?

> When reading from a pipe (e.g. nix-build | cachix push), the parent may be
> killed while uploads are in-flight. HTTP requests hang on dead connections and
> the process spins at 100% CPU.

How does the pipe change things? `nix-build` isn't the parent process in this case.

### 2. How do we handle broken connections?

This is the more serious issue that's not resolved here.

`responseTimeoutNone` is a red-herring. I don't think it comes into play during the body upload and should be set to something reasonable to protect against hangs during connection establishment.

For the body uploads, we'd need TCP keep-alive and/or an activity-based timeout.
The former requires modifying http-client-tls and the latter is doable on our end by hacking on the network part of the conduit.

### Thoughts

This PR changes the behavior of 1. which then has the side-effect of fixing 2., but only in the scenario where a pipe is used.
The big question is whether 1. needs changing at all.

danv-anduril:
Thanks - you're right that the parent watchdog is a band-aid and the real issue is dead connection detection - if connections didn't hang we wouldn't get into this 100% CPU usage scenario. Going to close this PR.

Here's the simplest repro for this issue (tested on 1.11.0):
```
cachix push <cache> /nix/store/<large-path>
sudo iptables -A OUTPUT -d <cachix-ip> -j REJECT
# cachix hangs indefinitely - no timeout, no retry, no exit
```

DROP instead of REJECT also reproduces the hang.

sandydoo:
I have code stashed somewhere to track network bytes. That'll get us most of the way I think.

Issue #425, opened 2022-07-28
Secret management


domenkozar:
Allow provisioning secrets to agents via the administrative console.

dzmitry-lahoda:
hm, could it be next:

1. agent provides his public key to dashboard
2. admin host ecrypts secrets for agent using that key (it can be shared key or per agent)
3. it puts encrypted into cachix/nix (there 3-4 techs i am aware do it, likely more - what best for nix)
4. agents obtain and decrypt.

it may decomposed with terraform, so terraform reads secrets from cloud and runs command to produce encrypted versions of secrets. 

`administrative console` sounds too GUI and not IaC.  

some nix glue to helo to set wiring up.

ideal solution for web3.0

PR #737, opened 2026-07-20
Resolve CACHIX_AUTH_TOKEN and CACHIX_SIGNING_KEY via secretspec

### LINKED ISSUE #89: Password manager support
Instead of storing secret key into ~/.config/cachix/cachix.dhall, Cachix could integrate with password managers. 

Requested in https://github.com/cachix/feedback/issues/3

domenkozar:
## Summary

Adds an optional integration with [secretspec](https://secretspec.dev) for resolving and storing cachix credentials, behind a new `secretspec` cabal flag.

### Resolution order

Each secret keeps its existing lookup order and gains secretspec as the final fallback, so a manifest can never override an existing setup and providers are only queried when cachix would otherwise have no credentials:

1. environment variable (`CACHIX_AUTH_TOKEN` / `CACHIX_SIGNING_KEY`)
2. `cachix.dhall` configuration file (per cache config for the signing key)
3. the project's `secretspec.toml`, walking up from the working directory
4. a manifest embedded in the binary with project name `cachix`

Providers namespace secrets by project name, so with step 4 a secret stored once under e.g. the keyring entry `secretspec/cachix/default/CACHIX_AUTH_TOKEN` works from any directory.

### Storing secrets

`cachix authtoken` and `cachix generate-keypair` store credentials via secretspec **by default** when it is usable on the machine (the `secretspec` CLI on PATH plus a configured default provider). When the CLI is installed but unconfigured, interactive runs offer to run `secretspec config init` on the spot and store via it immediately; non-interactive runs fall back to the configuration file with a tip. `--secretspec` forces the secretspec store, `--no-secretspec` keeps the configuration file. Storing via secretspec also moves the corresponding credential out of `cachix.dhall`, which takes precedence on reads and would otherwise shadow the new one. The embedded `cachix` namespace is pinned to the `default` profile on both store and resolve, so credentials stay reachable regardless of `SECRETSPEC_PROFILE` or the user's configured default profile.

A bare `cachix authtoken` delegates to the secretspec CLI's hidden prompt, keeping the token out of argv and shell history. Signing keys use a [secret reference](https://secretspec.dev/concepts/references/) per cache (`CACHIX_SIGNING_KEY_<CACHE>`, cache name folded to a valid env identifier), so each cache has its own slot on both store and resolve. Writes go through the `secretspec` CLI since the linked resolver is read only, passing an explicit `--reason` to satisfy the `require_reason` policy in agent environments unless the user exported `SECRETSPEC_REASON`.

### Build

The Haskell SDK statically links `libsecretspec_ffi.a`, which Hackage build bots and plain nixpkgs builds cannot provide, so the cabal flag defaults to off. The flake builds the archive from a pinned secretspec input (libdbus vendored in, so no new runtime dependencies, confirmed via ldd) and enables the flag on Linux. All secretspec logic lives in the new `Cachix.Client.SecretSpec` module, the single CPP site; without the flag it compiles to stubs.

### Verification

- token and signing key resolve from a project manifest, from the embedded namespace, and fall through cleanly when nothing is resolvable; env var and config file still win
- full round trip with the built binaries: `cachix authtoken --secretspec` stores through the secretspec CLI, `cachix doctor` reads it back with no env var and no config file
- flag off build, flag on build, and `nix build .#cachix` all pass; ormolu, cabal-fmt, hlint, nixfmt clean

### Known limitations

- darwin keeps the flag off in the flake until Security framework link flags for the Rust archive are wired up
- the FFI resolves the whole manifest at once, so a project manifest with other missing required secrets masks a resolvable cachix secret until secretspec grows a per secret request

### Issues

Closes #89: secretspec providers cover `pass`, gopass, keyring, 1Password, LastPass and more, for both resolving credentials and storing them via the `--secretspec` flags (note: Nix builds enable the feature on Linux only for now). Related but not addressed here: #425, which is about provisioning secrets to Cachix Deploy agents.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


alurm:
I locally rebased this PR onto current master and tested the result on aarch64-linux and aarch64-darwin.
  
  On aarch64-linux, the build enabled secretspec=True and passed all 113 tests, including resolution through a real dotenv provider.
  
  I also manually tested the pinned SecretSpec v0.16 CLI write path with cachix authtoken --stdin --secretspec, using an isolated fake token without modifying real credentials.
  
  The Linux pre-commit check passed: cabal-fmt, HLint, nixfmt, Ormolu, and ShellCheck.
  
  On aarch64-darwin, the package build passed all 108 applicable tests and the complete pre-commit suite passed. Cachix currently disables SecretSpec support on macOS, so the SecretSpec integration tests are not run there.
  
  Testing the rebased PR exposed several rebase and integration issues. I added one follow-up commit that:
  
  - fixes the signing-key lookup while preserving the last non-empty key from duplicate cache entries;
  - adds a missing test import;
  - aligns the setup command and access-reason handling with SecretSpec v0.16;
  - fixes a path collision in the fake SecretSpec CLI test;
  - corrects the git-hooks module attribute;
  - updates stale documentation; and
  - adds regression coverage for the affected behavior.
  
  https://github.com/alurm/cachix/tree/pr-737 contains the complete rebased PR followed by my additional fix commit.
  
  The follow-up fixes are contained in commit https://github.com/alurm/cachix/commit/6885b91d33fd5742590c28fedfb4325ee892bb5f, which can be cherry-picked after rebasing the original PR.

PR #738, opened 2026-07-20
nixconf: manage caches via an !include fragment (#413)

### LINKED ISSUE #413: `cachix use` overwrites globally configured substituters
I have a GH Actions workflow where I'm trying to configure a local binary cache restored from actions/cache, as that is quite a bit faster than always querying Cachix:
```
      - uses: cachix/install-nix-action@v15
        with:
          extra_nix_config: |
            extra-substituters = file://${{ github.workspace }}/nix-store-cache-copy?priority=10&trusted=true
      - uses: cachix/cachix-action@v10
        with: ...
```
However, despite the use of `extra-substituters`, it looks like Cachix' config in `~/.config/nix/nix.conf` completely replaces the above config:
```
+ cat /home/runner/.config/nix/nix.conf
substituters = https://cache.nixos.org https://lean4.cachix.or
trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= lean4.cachix.org-1:mawtxSxcaiWE24xCXXgh3qnvlTkyU7evRRnGeAhD4Wk=
+ nix show-config
...
substituters = https://cache.nixos.org https://lean4.cachix.org
```
Presumably this would be avoided if Cachix used `extra-substituters` as well.

domenkozar:
## What

`cachix use` no longer edits the contents of your `nix.conf`. It writes the caches it manages to a dedicated `cachix.conf` fragment next to the `nix.conf` it configures, and pulls that fragment in with a single `!include cachix.conf` line. From then on cachix only rewrites its own fragment and leaves the rest of your Nix settings untouched (fixes #413).

Concretely:

- **Caches live in a fragment cachix owns.** Substituters and public keys are written to `cachix.conf` as `extra-substituters` / `extra-trusted-public-keys`, which append to whatever Nix and other tools already configure rather than overriding it.
- **The user's `nix.conf` gets one stable line.** `!include cachix.conf` is added once if missing; it is the optional form, so deleting the fragment does not break Nix.
- **Upgrades migrate cleanly.** Any cache settings older versions wrote inline in the `nix.conf` (the old `substituters =` / `trusted-public-keys =` override lines, including the `cache.nixos.org` default they injected as a workaround) are lifted into the fragment and removed from `nix.conf`.

The upshot: cachix only manages the caches it was asked to add, in a file it fully owns. Removing a cache from another config (for example `substituters =` in the global `nix.conf`) now takes effect instead of being masked by a stale copy cachix pasted in.

Because `extra-*` and `!include` rely on newer Nix, the client now requires **Nix 2.4 or newer**; `cachix use` and `cachix remove` assert the version before touching config.

## Details

- `NixConf`: new `ExtraSubstituters` / `ExtraTrustedPublicKeys` constructors (parsed and rendered); `addCache` / `removeCache` operate on the `(nix.conf, fragment)` pair, building the fragment's `extra-*` settings and, on the `nix.conf` side, stripping cachix-managed cache lines and ensuring the `!include`. Empty assignments are omitted rather than written blank. Added `readPathWithDefault` and the `cachixConf` fragment name.
- `InstallationMode`: `addBinaryCache` / `removeBinaryCache` read the `nix.conf` and the fragment, write the fragment, and rewrite `nix.conf` only when it actually changes (so a settled config sees no churn). Private-cache `netrc-file` goes in the fragment too.
- `NixVersion`: `isSupportedNixVersion` gates on 2.4, handling both `Ideal` (3 component) and `General` (2 component, e.g. `2.4`) version strings.
- `remove` command asserts the Nix version, matching `use`.

## Testing

`stack test`: 115 examples, 0 failures. `addCache` / `removeCache` specs cover writing to the fragment and adding the include, accumulating across runs, migrating inline settings from older versions (and dropping the injected Nix defaults), leaving unrelated `nix.conf` settings in place, not adding a duplicate include, removing from the fragment, omitting empty assignments, and no-op removal leaving both configs untouched. Plus the 2.4 version boundary.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cachix/src/Cachix/Client/InstallationMode.hs
+++ b/cachix/src/Cachix/Client/InstallationMode.hs
@@ -4,6 +4,7 @@ module Cachix.Client.InstallationMode
   ( InstallationMode (..),
     NixEnv (..),
     getNixEnv,
+    requireNixEnv,
     getInstallationMode,
     addBinaryCache,
     removeBinaryCache,
@@ -16,11 +17,13 @@ module Cachix.Client.InstallationMode
   )
 where
 
+import Cachix.API.Error (escalateAs)
 import Cachix.Client.Config (Config)
 import Cachix.Client.Config qualified as Config
 import Cachix.Client.Exception (CachixException (..))
 import Cachix.Client.NetRc qualified as NetRc
 import Cachix.Client.NixConf qualified as NixConf
+import Cachix.Client.NixVersion (assertNixVersion)
 import Cachix.Client.URI qualified as URI
 import Cachix.Types.BinaryCache qualified as BinaryCache
 import Data.Maybe qualified
@@ -81,7 +84,12 @@ toString UntrustedNixOS = "untrusted-nixos"
 getNixEnv :: IO NixEnv
 getNixEnv = do
   user <- getUser
-  ncs <- NixConf.resolveIncludes =<< NixConf.readWithDefault NixConf.Global
+  -- The global nix.conf is only inspected for trusted users here, never
+  -- written back, so an unreadable file can safely degrade to an empty one.
+  globalConf <-
+    NixConf.read NixConf.Global
+      >>= maybe (NixConf.new <$> NixConf.getFilename NixConf.Global) pure
+  ncs <- NixConf.resolveIncludes globalConf
   isTrusted <- isTrustedUser $ concatMap (NixConf.readLines NixConf.isTrustedUsers) ncs
   isNixOS <- doesFileExist "/run/current-system/nixos-version"
   return $
@@ -90,6 +98,13 @@ getNixEnv = do
         isTrusted = isTrusted,
         isNixOS = isNixOS
       }
+
+-- | Assert the installed Nix is new enough, then resolve the environment
+-- both `cachix use` and `cachix remove` need before touching nix.conf.
+requireNixEnv :: IO NixEnv
+requireNixEnv = do
+  () <- escalateAs UnsupportedNixVersion =<< assertNixVersion
+  getNixEnv
 
 getInstallationMode :: NixEnv -> UseOptions -> InstallationMode
 getInstallationMode nixenv useOptions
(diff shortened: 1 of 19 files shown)


cachix/src/Cachix/Client/InstallationMode.hs
```diff
@@ -136,73 +151,228 @@ b) Run the following command to add your user as trusted
 |]
 addBinaryCache config bc useOptions WriteNixOS =
   nixosBinaryCache config bc useOptions
+-- A custom output directory (--output-directory) is a fully cachix-generated
+-- artifact meant to be shipped elsewhere, so its nix.conf stays
+-- self-contained instead of splitting the caches into a fragment.
+addBinaryCache config bc _ (Install ncl@(NixConf.Custom _)) = do
+  nixConf <- NixConf.readWithDefault ncl
```
alurm, line 158:
```suggestion
  nixConf <- NixConf.readPathQuiet =<< NixConf.getFilename ncl
```

cachix/src/Cachix/Client/InstallationMode.hs
```diff
@@ -136,73 +151,228 @@ b) Run the following command to add your user as trusted
...
+  !include ${NixConf.cachixConf}
+|]
+  printMigrationNotice migrate nixConf fragment
+  _ <- writeWithHint fragmentWriteHint fragment (addNetRCLine (fragment {NixConf.nixConfLines = fragment'}))
+  _ <- writeWithHint includeHint nixConf (clearStaleNetRCLine (nixConf {NixConf.nixConfLines = nixConf'}))
+  putStrLn $ "Configured " <> BinaryCache.uri bc <> " binary cache in " <> fragmentPathT
 
 removeBinaryCache :: URI.URI -> Text -> InstallationMode -> IO ()
+removeBinaryCache uri name (Install ncl@(NixConf.Custom _)) = do
+  nixConf <- NixConf.readWithDefault ncl
```
alurm, line 205:
```suggestion
  nixConf <- NixConf.readPathQuiet =<< NixConf.getFilename ncl
```

cachix/src/Cachix/Client/InstallationMode.hs
```diff
@@ -136,73 +151,228 @@ b) Run the following command to add your user as trusted
...
+  where
+    stripSlash = T.dropWhileEnd (== '/')
+
+-- | Resolve the nix.conf at the given location and the cachix.conf fragment
+-- it includes, reading both from disk.
+resolveNixConfAndFragment :: NixConf.NixConfLoc -> IO (NixConf.NixConfSource, NixConf.NixConfSource)
+resolveNixConfAndFragment ncl = do
+  nixConfPath <- NixConf.getFilename ncl
+  let fragmentPath = replaceFileName nixConfPath (toS NixConf.cachixConf)
+  nixConf <- NixConf.readPathWithDefault nixConfPath
```
alurm, line 277:
```suggestion
  nixConf <- NixConf.readPathQuiet nixConfPath
```

